PR #40, opened 2015-02-13
[fix] Process Channel Errors regardless of status

alex-deaconu:
Split out Channel Error in its own function
Process error even in inactive state

Fixed typo with hasData flag: if Viewer has data it should stop polling

Updated default poll and ping interval to 30 minutes


alex-deaconu:
@stevesherrie couple of fixes for the Viewer logic; have passed QA and are pending release. 

Please review. Thanks!

stevesherrie:
:+1:

PR #45, opened 2015-05-10
Update Viewer.html

woneal1983:
I would think that the script would go there since its producing a meta tag which would be placing it in the correct spot. It probably doesn't matter but for standard and organization i figured it would need to be in the correct spot. :)


alex-deaconu:
Replaced by #46

PR #47, opened 2015-07-13
Add viewport by default; set width/height if parameters are given

alex-deaconu:
Also if viewport=false parameter is present, don't add anything

Original Commit Message from woneal1983:
I can do without the viewport tag all together on my amazon and google tv apps, but for the new chromium 43 library for some reason has to have the specific height and width in the viewport in order to fix a scaling issue on boxes and sticks. Works right on tablets but not on boxes or sticks. So if you don't want any viewport at all then this new pull request will do what is needed. I also changed the parameters to vh and vw which works better for me. Anyway take a look and see if this fits.

### DIFF
--- a/src/com/risevision/viewer/client/info/Global.java
+++ b/src/com/risevision/viewer/client/info/Global.java
@@ -7,7 +7,7 @@ package com.risevision.viewer.client.info;
 import java.util.Date;
 
 public class Global {
-	public static final String VIEWER_VERSION = "1-06-038";
+	public static final String VIEWER_VERSION = "1-06-039";
 	
 //	public static final String GADGET_SERVER_URL = "http://rvagadgets.appspot.com/";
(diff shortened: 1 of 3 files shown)


alex-deaconu:
I've opened a new PR for this and finalized the commit. Deployed and tested to http://1-06-039.rvashow2.appspot.com/, and this is ready for review. 

@tejohnso @fjvallarino please do so when you have a moment.

@woneal1983 FYI.

Thanks.

tejohnso:
:+1:

woneal1983:
This is good to go for me. :)

alex-deaconu:
@woneal1983 this is now deployed. Thanks!

woneal1983:
Thanks Alex for helping out. Really Appreciate the help you all have done. Thanks

PR #48, opened 2015-07-30
Set up external logging

tejohnso:
Adds external logging functionality.  Can be tested by loading ext-logger-test.html in a browser.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.class
+*.swp
 
 # Package Files #
 # *.jar #
--- a/src/com/risevision/viewer/client/ViewerEntryPoint.java
+++ b/src/com/risevision/viewer/client/ViewerEntryPoint.java
@@ -71,6 +71,8 @@ public class ViewerEntryPoint implements EntryPoint {
 		
 //		showBlackScreen(true);
 		
+		ViewerHtmlUtils.logExternalMessage("startup");
+		
 		//check if no Display ID 
 		if (RisePlayerController.getIsActive() &&  isDisplay() && !isEmbed() && DEMO_ID.equals(id)) {
 			if (claimId != null && !claimId.isEmpty()) {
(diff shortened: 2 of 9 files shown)


alex-deaconu:
@tejohnso the Viewer is a public project. Is it a good idea to keep the oauth token/secret for the BQ project in here? Thanks.

tejohnso:
@alex-deaconu It's akin to storing an api token similar to segment or rollbar.  It's only scoped to insert into that project.  I'm not sure of another way for a browser client to be in control of the inserts without it knowing the credentials, unless we introduce some backend process.

alex-deaconu:
@tejohnso okay, thanks for the clarification. 

To confirm, you want me to deploy this to the Viewer server such that you can test the messaging functionality? Or you want me to integrate this with specific events in the Viewer? Thanks.

tejohnso:
@alex-deaconu I think for now we can deploy into viewer, and maybe get one event logging just to make sure it's working.  Maybe just a "startup" event for now.

alex-deaconu:
@tejohnso I've added a sample Viewer Started event and the helper function that will send events from GWT. This is deployed at http://1-06-040.rvashow2.appspot.com/. Tested and the event works. Thanks.

tejohnso:
Thanks @alex-deaconu   
Can you check out this change I just made?  I think just having these two events would be a pretty useful deployment.

alex-deaconu:
@tejohnso to confirm, we're logging the "started" event for all Viewers (including preview). Do you think it's a good idea to only log for Displays, or is that ok? Thanks.

tejohnso:
@alex-deaconu I think we should log both cases, but I don't like the idea of not being able to determine one from the other.  Maybe we should add a "type" field into the event database and include that in each record.

alex-deaconu:
@tejohnso sure. I can add a type field. To confirm, if the Viewer is in preview mode the Display Id would be empty. Thanks.

alex-deaconu:
@tejohnso I've added a try/catch just in case, and updated the linker (for html5 caching) with the script file. This is ready for merge. Please review. Thanks!

tejohnso:
:+1:

PR #50, opened 2015-08-04
New common.jar to fix JSON boolean parsing

alex-deaconu:
This fix is required to fix parsing of schedule items saved via the Schedules App.

@olegrise please review. @tejohnso @fjvallarino FYI. Thanks!


olegrise:
:+1:

PR #51, opened 2015-08-05
Add continuous integration

fjvallarino:
This PR moves the project to Maven builds and CircleCI integration. These are the most important changes:

- Source code and resource files were moved to follow Maven's file structure
- Version number follows the "1-6-" format, but adds a YYYYMMDDhhmm number instead of a sequence.
- Environment selection is handled on the CircleCI build, which updates the corresponding Global.java file. It defaults to staging values for local development.
- Twitter credentials are handled with CircleCI environment variables and source code replacement.
- Integration with Eclipse is not currently working, since WTP does not copy the generated viewer.nocache.js (and related) to the corresponding location. I will investigate this and submit a new PR as soon as I find a solution. The server works properly when running it from command line.

@alex-deaconu @tejohnso please review


alex-deaconu:
@fjvallarino nice job on this. It looks good to me, although I'm quite new at maven. :+1: .

@ByronDarlison I'm wondering if we should do the same thing with the RVA. It's probably 1 days' work but the benefit is we have CI for that project as well. Just a thought. Thanks.

alex-deaconu:
Oh, almost forgot. @fjvallarino please don't forget to squash before merge. Thanks!

ByronDarlison:
@alex-deaconu let's hold off for now and hopefully we can eliminate the problem by eliminating RVA ;-)

tejohnso:
Looks good.  
So when this merges it's going to push out a new viewer version and make it default.  Did we get Blake to smoke test maven-ci-dot-rvaviewer-test.appspot.com?

PR #54, opened 2015-08-07
Run export and awk command in a single shell line to overcome "one shell per command" CircleCI policy

fjvallarino:
@tejohnso please review


tejohnso:
:+1:

PR #58, opened 2015-08-20
Delay core api and channel calls on startup

tejohnso:
This will attempt to retrieve previously saved api data.  If there is no previously saved data then the api call will be made.


fjvallarino:
:+1:

PR #62, opened 2015-09-02
Fix/metarefresh

tejohnso:
Meta refresh causes serious problems in presentations.
See https://code.google.com/p/chromium/issues/detail?id=527367


fjvallarino:
:+1:

PR #63, opened 2015-09-03
Improve Eclipse integration

fjvallarino:
These changes are needed to be able to properly load the project on Eclipse. It does not yet fix the issue of running and debugging inside the IDE.

@tejohnso please review


tejohnso:
:+1:

PR #64, opened 2015-09-03
Ignore update_required messages

fjvallarino:
This ignores the update_required by always providing empty string as the value. Probably removing the line I commented out would be better.

@tejohnso please review


tejohnso:
:+1:

PR #82, opened 2016-01-25
Fix empty playlist operation

alex-deaconu:
Don't send Done if there are no items

### DIFF
--- a/src/main/java/com/risevision/viewer/client/controller/ViewerPlaceholderController.java
+++ b/src/main/java/com/risevision/viewer/client/controller/ViewerPlaceholderController.java
@@ -323,7 +323,7 @@ public class ViewerPlaceholderController {
 			}
 		}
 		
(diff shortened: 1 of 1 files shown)


src/main/java/com/risevision/viewer/client/controller/ViewerPlaceholderController.java
```diff
@@ -323,7 +323,7 @@ public boolean play() {
 			}
 		}
 		
-		return isPlaying || placeholderDoneCommand == null;
+		return gadgets.size() == 0 || isPlaying || placeholderDoneCommand == null;
```
alex-deaconu: Returning `true` prevents PUD from being called for the Presentation.

alex-deaconu:
@tejohnso @ezequielc please review. Thanks!

ezequielc:
@alex-deaconu :+1:

PR #84, opened 2016-01-28
non-Visible placeholders don't send done

alex-deaconu:
Fixing issue where non-Visible placeholder sends Done, receives Play, sends Done, infinitely. Very similar to the 0 item placeholder issue fixed in the previous PR.

@ezequielc @tejohnso please review. Thanks.


ezequielc:
@alex-deaconu :+1:

PR #89, opened 2016-03-09
Log Display Activated event

ezequielc:
- Added `Display Activated` event tracking to BQ and GA
- Update our deploy authentication as Google deprecated the old API we were using.

I think we should queue a card to use BQ Template Tables in  `logExternal()` method as well.

@alex-deaconu  Please review. Thanks!

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,3 +1,7 @@
+dependencies:
+  post:
+    - git clone git@github.com:Rise-Vision/private-keys.git
+    - cp private-keys/core/.appcfg_oauth2_tokens_java ~
 checkout:
   post:
     # Update environment depending URLs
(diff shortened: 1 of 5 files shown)


alex-deaconu:
@ezequielc looks good :+1: . Tested with a Display and the event was logged in BQ.

Curious, what advantages does using the Template Tables give us?

Thanks.

ezequielc:
@alex-deaconu It creates our daily tables automatically based on the suffix. Before that functionality was provided by BQ, we had a service to create our tables every day: https://github.com/tejohnso/bigquery-table-inserter

Thanks!

alex-deaconu:
@ezequielc cool, thanks for the clarification. Agreed re: card. Thanks.

PR #90, opened 2016-03-31
Add re-try timer for Rise Cache

alex-deaconu:
Timer only executes if Cache fails to respond
Log Cache response failures to BQ

Log non widget item initialization

### DIFF
--- a/src/main/java/com/risevision/viewer/client/cache/RiseCacheController.java
+++ b/src/main/java/com/risevision/viewer/client/cache/RiseCacheController.java
@@ -5,11 +5,14 @@
 package com.risevision.viewer.client.cache;
 
 import com.google.gwt.http.client.URL;
+import com.google.gwt.user.client.Timer;
 import com.risevision.common.client.utils.RiseUtils;
+import com.risevision.viewer.client.player.RisePlayerController;
+import com.risevision.viewer.client.utils.ViewerHtmlUtils;
 
 public class RiseCacheController {
 //	private static RiseCacheUtils instance;
-	private static boolean isActive = false; 
+	private static boolean isActive = false;
 	
 //	public RiseCacheUtils() {
 //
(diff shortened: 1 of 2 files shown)


alex-deaconu:
@rodrigopavezi this PR creates a retry mechanism and enables logging for Rise Cache. Hopefully it will help you identify the issue where cache is not present in some situations. @ezequielc FYI.

Please review. Thanks.

src/main/java/com/risevision/viewer/client/cache/RiseCacheController.java
```diff
@@ -59,10 +70,22 @@ public static void pingCache() {
 		if (!isActive) {
 			String url = "http://localhost:9494/ping?callback=?";
 			pingCacheNative(url);
+			
+			if (RisePlayerController.getIsActive()) {
```
rodrigopavezi: @alex-deaconu shouldn't it be if not is active here? cheers

alex-deaconu: @rodrigopavezi actually, this checks that the Viewer is running on a Player (Rise or 3rd party) rather than simply a URL in the browser. The code 3 lines above `if (!isActive)` checks if the Rise Cache response was received.

Thanks.

rodrigopavezi:
@alex-deaconu looks good :+1: thanks for the clarification.

alex-deaconu:
@rodrigopavezi quick question, is this change consistent with the tracking for Widget usage that Content has implemented. And, is the Rise Cache logging consistent with the logging that Content has implemented recently for Rise Cache not being active?

Thanks!

rodrigopavezi:
@alex-deaconu I think it does. We log every time a widget goes into the play and done state for usage. Then we also log error events when they happen such as the rise cache ping failure. Actually you have implemented it better with the count of the pings. Moreover we haven't implemented the retry on the storage component. Something we have to do soon and I will mention about the ping counter. Cheers

alex-deaconu:
@rodrigopavezi sounds good. Thanks for the clarifications!

PR #96, opened 2016-05-04
No longer send Done for PH Timeline failure

alex-deaconu:
I've modified the Presentation used in the previous validation to be used for this as well.

To clarify, a Play Until Done placeholder is set to not play in the Placeholder timeline, however the Placeholder should have items in it. Previously (on http://rvaviewer-test.appspot.com) it should generate a stack error. On http://stack-error.rvaviewer-test.appspot.com/ that error should be gone.

@ezequielc please review. Thanks.


ezequielc:
@alex-deaconu :+1: Cheers!

PR #98, opened 2016-06-29
Check for null or empty duration string

alex-deaconu:
Fixes an issue in the Viewer were the `duration` is `null`. Still investigating the reason for the `null` value; however this will allow Item playback.

@ezequielc please review. Thanks.


ezequielc:
@alex-deaconu 👍

PR #99, opened 2016-07-11
Detect and add logging for done event loop

alex-deaconu:
If a Widget sends done more than once a second, record it
If it has done it 10 times in a row, log to BQ
Ignore it after

Adding console entry for BQ logging

### DIFF
--- a/src/main/java/com/risevision/viewer/client/controller/ViewerWidgetController.java
+++ b/src/main/java/com/risevision/viewer/client/controller/ViewerWidgetController.java
@@ -4,6 +4,8 @@
 
 package com.risevision.viewer.client.controller;
 
+import java.util.Date;
+
 import com.google.gwt.http.client.URL;
 import com.google.gwt.user.client.Command;
 import com.google.gwt.user.client.Window;
@@ -23,6 +25,9 @@ public class ViewerWidgetController implements ViewerControllerInterface {
 //	private boolean canPlay, canStop, canPause, canReportReady, canReportDone;
 	private Command gadgetReadyCommand;
 	private Command gadgetDoneCommand;
+	
+	private long lastDone = 0;
+	private int doneCounter = 0;
 
 	private static final String WIDGET_SCRIPT = "<script><!--\n" +
 			"try {" +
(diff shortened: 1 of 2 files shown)


alex-deaconu:
Logging to BQ if the loop happens more than 10 times for a particular widget instance. Also, not responding to the done event more than once a second.

To clarify, this doesn't resolve the issue with the Facebook Widget, but at least it will allow us to detect this incorrect behaviour.

Replicate at (takes about 10-20 mins for the event to be logged): http://done-loop-dot-rvaviewer-test.appspot.com/?type=schedule&id=6d3b9c16-c3f4-4bcc-8c34-17d782dbd997&showui=false

Query to validate: `select * FROM [Viewer_Events.events20160711] WHERE event = 'Widget Done Loop' order by ts desc`

@ezequielc please review. @tejohnso FYI. Thanks!

circle.yml
```diff
@@ -5,7 +5,7 @@ dependencies:
 checkout:
   post:
     # Update environment depending URLs
-    - if [ $CIRCLE_BRANCH == "master" ]; then export DEPLOYENV=prod; else export DEPLOYENV=staging; fi; cat src/main/java/com/risevision/viewer/client/info/Global.java | awk -v DEPLOYENV=$DEPLOYENV 'BEGIN{OFS=FS="="}/String ENV/{$2=" \"" DEPLOYENV "\";"}1' | awk 'gsub("{dateTime}", strftime("%Y%m%d%H%M"))1' > Global.tmp
+    - if [ $CIRCLE_BRANCH == "master" ]; then export DEPLOYENV=prod; else export DEPLOYENV=prod; fi; cat src/main/java/com/risevision/viewer/client/info/Global.java | awk -v DEPLOYENV=$DEPLOYENV 'BEGIN{OFS=FS="="}/String ENV/{$2=" \"" DEPLOYENV "\";"}1' | awk 'gsub("{dateTime}", strftime("%Y%m%d%H%M"))1' > Global.tmp
```
ezequielc: @alex-deaconu Curious, why did you need to change the environment here? Thanks

ezequielc:
@alex-deaconu Code looks good. Just one question. Thanks

src/main/java/com/risevision/viewer/client/controller/ViewerWidgetController.java
```diff
@@ -153,6 +158,24 @@ public void setError(String presFrame, String reason) {
...
+			if (lastDone != 0 && new Date().getTime() - lastDone < 1000) {				
+				lastDone = new Date().getTime();
+
+				// -1 indicates already logged
+				if (doneCounter != -1) {
+					doneCounter++;
+					
+					if (doneCounter >= 10) {
+						ViewerHtmlUtils.logExternalMessage("Widget Done Loop", playlistItem.getObjectRef());
+
```
tejohnso: @alex-deaconu Can you change to "widget done loop" for consistency?

alex-deaconu: @tejohnso corrected. Thanks for the heads up!

circle.yml
alex-deaconu: @ezequielc @tejohnso sorry forgot to mention this. The Schedule I used for testing works only in Production (the Widget is not registered on the Test server), so I changed the staging to `prod` for testing.

I've just reverted the change however. Thanks.

tejohnso:
:+1:

ezequielc:
@alex-deaconu 👍

PR #104, opened 2016-08-12
Report Viewer version to Player

fjvallarino:
The postToPlayer function is injected by Electron Player. I was planning on using an if clause to check for that function's existence, but since all the other native functions already included a try/catch clause, I followed the same pattern.

@alex-deaconu @tejohnso please review


tejohnso:
:+1:

alex-deaconu:
@fjvallarino @tejohnso looks good 👍 . Thanks.

PR #108, opened 2016-08-30
Listen for content updated over messaging

ahmedalsudani:
@alex-deaconu @fjvallarino @tejohnso ready for review.

This change makes it possible to update Viewer contents through messaging.

It's a fairly simple change if you ignore all the debugging code. Only the first commit (76111b0) includes code that would be merged.

If you want to test integration with launcher, you can use [this branch of launcher](https://github.com/Rise-Vision/rise-launcher-electron/tree/staging/test-viewer-listens-for-content-updated). It won't update the presentation, but the console shows the message being propagated to `reportDataReady`

A corresponding change in the launcher repo is in progress.


tejohnso:
@alex-deaconu This looks ok to me unless you wanted to take this opportunity to discuss and alternative entry point and avoid GWT code here.

alex-deaconu:
@tejohnso @ahmedalsudani just a comment on the PR. The Viewer will get its own data - and render it, and that data will be overlapped by the data sent from the Player. I think it's ok for prototyping purposes, just want to clarify that's not the intended final behaviour. Which also begs the question of whether or not this code should make it to production or you can just use a branch for testing/validation (I'm ok either way).

Also, Ahmed we're trying to scope out GWT code as much as possible, so I'm wondering if it's possible to move this functionality to Javascript. Ideally, just surface a function here: 
https://github.com/Rise-Vision/viewer/blob/master/src/main/java/com/risevision/viewer/client/utils/ViewerHtmlUtils.java#L137
that will be called from your JS functionality. Let me know if that makes sense.

Thanks!

tejohnso:
@alex-deaconu Updated to remove GWT code and only log the retrieved content for now.  Once we see some good log data we can move to replace the current channel functionality.

alex-deaconu:
@tejohnso sounds good 👍 . Wondering if you could squash before merging. Thanks!

PR #112, opened 2016-09-12
Chore/jsdom experiment

settinghead:


settinghead:
@alex-deaconu @ezequielc This PR switches unit test environment from Karma + Mocha + PhantomJS to purely Node.js + Mocha (with JSDom as simulated browser environment).

Pros:
1. PhantomJS & Karma are no longer needed.
2. Only nodejs is require to run all unit tests. 
3. Tests for each module are largely isolated to the module itself. No need to populate all namespaces to test a module
4. Each time a test is run, the simulated DOM structure resets itself to a fresh state
5. Can now test with fake URL & `window.location`
6. No longer need dependency injection style of code
7. Faster unit test executions

Cons:
1. Browser environment is now simulated, whose behaviour might be slightly different from Webkit/Gecko/Edge browsers) in some cases

Please take a look and let me know if you like this approach and whether we can adopt it. Thanks!

src-js/Viewer/Utils.tests.js
```diff
@@ -1,84 +1,99 @@
...
   });
-  
-  xit('getFromQueryString: ', function() {
-    var mockedLocation = {
-      search: '?type=display&id=1d3b23d3-31f6-4ad8-b8b1-c180231b9919'        
-    };
-    window.location = mockedLocation;
+
+  it('getFromQueryString: ', function() {
+    jsdom.changeURL(window, 'https://lolapps.risevision.com/?type=display&id=1d3b23d3-31f6-4ad8-b8b1-c180231b9919');
```
settinghead: @alex-deaconu browser URL simulation ^

settinghead:
Not useful right now. Closing.

PR #116, opened 2016-09-13
Channel connection refactor

alex-deaconu:
Fix dataController.tests
Increment viewer version
Some lint fixes


alex-deaconu:
@ezequielc @tejohnso @settinghead please review. Thanks!

src-js/Viewer/Channel/ChannelConnectionController.tests.js
```diff
@@ -11,39 +66,140 @@ describe('ChannelConnectionController: ', function() {
...
+    var channelDisconnectedStub = sinon.stub(RiseVision.Viewer.Channel.ChannelConnectionProvider, 'channelDisconnected', function() {});
+
+    service.connectionCancelled();
     
-    RiseVision.Viewer.Channel.ChannelConnectionController.connectionCancelled();
+    expect(service.isInactive()).to.be.false;
+    channelDisconnectedStub.should.have.been.called;
     
-    jsniConnectionCancelledSpy.should.have.been.called;
+    channelDisconnectedStub.restore();
```
settinghead: Ideally stubbing/spying and restoring should happen in `beforeEach` and `afterEach` blocks, because if say this test fails, the stub/spy will not be properly restored, which might affect the execution of subsequent tests.

src-js/Viewer/Channel/ChannelConnectionProvider.js
```diff
@@ -0,0 +1,148 @@
...
+  CONNECT: 'connect',
+  ERROR: 'error',
+  CLOSE: 'close',
+  UPDATED: 'updated',
+  PING: 'ping',
+  AYT: 'ayt',
+};
+
+var HTML_STRING = '' +
+	'<html>\n' +
```
settinghead: FYI now that ES2015 is enabled, you could also use template literals:

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals

settinghead:
Other than the comment above re. where to stub/spy, LGTM.

src-js/Viewer/Channel/ChannelConnectionController.tests.js
alex-deaconu: @settinghead good point. Will fix. Thanks.

alex-deaconu:
@settinghead fixed the issue with the `sinon.stub` initialization by moving to beforeEach and afterEach functions.

Also converted the String as suggested.

In addition, added an extra check for the error code value `onChannelError`. It seems the value comes as a String. Not sure if this was always the case or it's new - so I left the existing integer check as well.

Please review. Thanks!

settinghead:
@alex-deaconu 

I noticed there are 2 more places that has `sinon.stub()` in `it()`, then realised I have been making the same mistakes myself so there are probably more in some other tests.

I suggest we leave them for now and come back and fix them in one go at a later point. What do you think?

Otherwise looks good.

ezequielc:
@alex-deaconu I might have found one issue. When I try to use an invalid Display ID, I don't get the invalid ID error message. The spinner keeps running and in the log I can see a `channel error` event. Could you try it as well? Thanks

src-js/Viewer/Data/ViewerDataController.js
```diff
@@ -40,110 +40,107 @@ var _pollingTimerExecute = () => {
...
+
+    break;
+  case Global.VIEWER_STATUS.DISPLAY_NOT_FOUND:
+    _showDisplayRegistration(NotificationType.display_id_not_found);
+
+    break;
+  default:
+    break;
+  }
+
```
ezequielc: @alex-deaconu In the java code there was a `updateViewerData(newViewerData);` here. Did you remove it intentionally? Thanks!

alex-deaconu:
@ezequielc yes, I'm aware of this issue. It's due to missing `NotificationType` functionality. See PR associated to this branch for fix. Thanks.

src-js/Viewer/Data/ViewerDataController.js
alex-deaconu: @ezequielc kind of. That functionality is still in the GWT code and gets triggered when the  `window.jsniReportDataReady(result);` gets called. However, I realize I should probably have moved that to JS as well. Will do so when the Notification & Display Registration widgets are made available. Thanks.

ezequielc:
@alex-deaconu 👍

PR #119, opened 2016-09-15
e2e setup

settinghead:


settinghead:
@alex-deaconu @ezequielc E2E is set up in this PR based on Protractor, https://github.com/Rise-Vision/widget-tester and https://github.com/Rise-Vision/rv-common-e2e. It also comes with a trivial test case that follows the same format with what's in https://github.com/Rise-Vision/rise-vision-apps. Please review and let me know if you have any comments. Thanks!

circle.yml
```diff
@@ -8,6 +8,7 @@ dependencies:
   post:
     - git clone git@github.com:Rise-Vision/private-keys.git
     - cp private-keys/core/.appcfg_oauth2_tokens_java ~
+    - mvn test
```
ezequielc: @settinghead Curious, why do we need `mvn test` here? Also, I don't think it runs any test actually. Thanks

gulpfile.babel.js
```diff
@@ -40,14 +42,55 @@ gulp.task('watch', watchJS());
...
       .pipe(uglify())
       .pipe(rename({ suffix: '.min' })
       .pipe(gulp.dest(DIST_DIR))));
 
+gulp.task('test:e2e:core', ['test:webdrive_update'],factory.testE2EAngular({
+  browser: 'chrome',
+  loginUser: process.env.E2E_USER,
+  loginPass: process.env.E2E_PASS,
+  testFiles: function(){
+    try{
```
ezequielc: @settinghead This `try/catch` block is a workaround to run tests in parallel. In order to work we need an extra setup, but I don't think it is necessary right now. I think you should remove and just use `process.env.TEST_FILES`. Thanks

src/main/webapp/Viewer.html
```diff
@@ -77,7 +77,7 @@
       <p>&nbsp;</p>
     </div>
  	</div>
-	<div id="mainDiv" style="position:absolute; height: 100%; width: 100%; top:0px; left:0px;">
+	<div id="mainDiv" style="position:absolute; height: 100%; width: 100%; top:0px; left:0px;" class="e2e--main-container">
```
ezequielc: @settinghead As we have `id` here, I think it is not necessary to add the e2e class. But I don't have a strong opinion. Cheers

circle.yml
settinghead: @ezequielc Right, this command itself is a noop and doesn't do much, but this is mainly for caching. I noticed Maven repeatedly downloads a lot of jar files every time the test and deployment is run, causing delays and less readable output files. Adding this line will trigger downloading of dependencies before CI saves them to cache. It's not a complete solution but yesterday when I checked the amount of undesirable output seemed to have reduced for E2E tests.

gulpfile.babel.js
settinghead: OK will do. Thanks.

src/main/webapp/Viewer.html
settinghead: I prefer to keep it as it makes it clear that there is a test case associated with this element, so developers who wish to make future changes would hopefully be more careful. I know it's a bit bloated, but I think it's a good separation of concern to have. Thanks.

gulpfile.babel.js
settinghead: @ezequielc updated and removed the try catch block. Thanks.

circle.yml
settinghead: @ezequielc I added a comment to clarify.

ezequielc: @settinghead Thanks!

gulpfile.babel.js
ezequielc: @settinghead I think you don't need the function() and return statement, only `testFiles: process.env.TEST_FILES` should work. Could you confirm? Thanks!

src/main/webapp/Viewer.html
ezequielc: @settinghead Sounds good to me. Thanks!

PR #122, opened 2016-09-16
Fix sysInfo logic

alex-deaconu:
@ezequielc @settinghead discovered this blunder while testing. Fixes reporting versions to the Core & Restart/Reboot. Please review. Thanks!


settinghead:
@alex-deaconu :+1:

PR #123, opened 2016-09-16
Player controller implementation

alex-deaconu:
@ezequielc @settinghead please review. Thanks!


src-js/Viewer/Player/RisePlayerController.js
```diff
@@ -0,0 +1,156 @@
...
+  }
+};
+  
+var _pingPlayer = function () {
+  if (isActive) {
+    var url = baseUrl + cmdHeartbeat + "?callback=?";
+    _callUrlNative(url, "ping");
+    _pingV3Watchdog();
+
+    setTimeout(_pingPlayer, HEARTBEAT_TIME);
```
settinghead: This is related to the comment above. I think having a global `setInterval` would suffice. If for some reason _pingPlayer() needs to be called on an ad-hoc basis (say as soon as the player is set to 'active'), it should just execute the function which also ensures one and only one execution loop is present. So maybe something like this:

```js
var pingLoop;

var _pingPlayer = () => {
  if (isActive) {
     ... 
  }
  if(!pingLoop) {
    pingLoop = setInterval(_pingPlayer, HEARTBEAT_TIME);
  }
};
```

alex-deaconu:
@settinghead good call. Will fix. Thanks!

src-js/Viewer/Player/RisePlayerController.js
alex-deaconu: @settinghead updated to use the suggested method. Please see second commit. Thanks!

settinghead:
@alex-deaconu all good now. :shipit:

ezequielc:
@alex-deaconu 👍

PR #124, opened 2016-09-19
Add `config` task to build

alex-deaconu:
[Note] This breaks the `npm start` task as watches
trigger in an endless loop when a changes is made
due to the config file being updated


alex-deaconu:
Quick fix. But please note this will break the task as described above. 

@ezequielc @settinghead please review. Thanks!

ezequielc:
@alex-deaconu 👍

PR #125, opened 2016-09-19
fix gulp dev infinite file watch loop issue caused by rewriting of Config.js

settinghead:


settinghead:
@alex-deaconu @ezequielc a quick fix, not smart but it should work.

alex-deaconu:
@settinghead 👍

ezequielc:
@settinghead 👍

PR #127, opened 2016-09-20
Get viewer data from player messages not core

tejohnso:
@alex-deaconu This prevents any use of the viewer data api call.   Still doing some testing.


tejohnso:
@alex-deaconu Updated using an explicit flag.

settinghead:
@tejohnso :+1:

src-js/Viewer/Data/ViewerDataProvider.js
```diff
@@ -76,6 +76,8 @@ var _dataTimerExecute = function() {
 };
 
 service.retrieveData = function(reason) {
+  if (window.disableViewerContentFetch) {return;}
+
```
alex-deaconu: @tejohnso wondering if it's better to move this check and below functionality (`window.receiveFromPlayer`) to the DataController `init()` (https://github.com/Rise-Vision/viewer/blob/master/src-js/Viewer/Data/ViewerDataController.js#L22). In that `if` statement we configure the data retrieval for the API vs. Embedded Presentations, so I think it would be a good fit. Thanks!

tejohnso: @alex-deaconu Looking at https://github.com/Rise-Vision/viewer/blob/master/src-js/Viewer/Data/ViewerDataController.js#L22 I don't see it configuring the data retrieval. 

It looks to me like it's going to initialize the ViewerInstanceController and then start an initial data retrieval. But we need to make sure that all data retrievals are bypassed including those from channel content update messages, or any other future call into retrieveData.

alex-deaconu: @tejohnso okay.

PR #128, opened 2016-09-21
Feature/error message

settinghead:


settinghead:
@ezequielc @alex-deaconu please review. Thanks.

src-js/Viewer/Data/ViewerDataController.js
```diff
@@ -72,40 +78,37 @@ service.reportDataReady = (result) => {
...
   case Global.VIEWER_STATUS.CONTENT_NOT_FOUND:
+  //TODO
     // _showNotification(newViewerData.getStatusMessage());
-    // 
-    // break;
+    _showNotification(newViewerData.status.message);
+    break;
+
   case Global.VIEWER_STATUS.ID_SHARING_VIOLATION:
-    // _showDisplayRegistration(NotificationType.display_id_duplicate);
```
alex-deaconu: @settinghead as mentioned in the GWT code, we still need to `_showDisplayRegistration` for this and the `DISPLAY_NOT_FOUND` messages. Thanks.

src-js/Viewer/Widget/ViewerNotificationsPanel.js
```diff
@@ -0,0 +1,47 @@
...
+    msg = notification;
+  } else {
+    msg = notification.getMessage();
+  }
+
+  return msg;
+}
+
+ViewerNotificationsPanel.getInstance = () => {
+  if(!instance) {
```
alex-deaconu: @settinghead I don't think we need an instance method for this. Would it be easier if we just had a static class? Thanks.

alex-deaconu:
@settinghead the `ViewerNotificationsPanel` is being used here (https://github.com/Rise-Vision/viewer/blob/master/src/main/java/com/risevision/viewer/client/controller/ViewerScheduleController.java#L139) in the GWT code. Wondering if you can remove/clean up the GWT code associated with the panel, and update this reference to use the GWT version (add a `native` method that opens the panel in GWT). Thanks.

src-js/Viewer/Widget/ViewerNotificationsPanel.js
settinghead: @alex-deaconu this is intended to make the loading asynchronous and separate it from initial DOM and namespace population. By making the initialization lazy (i.e. get instance only when the program needs it, which is long after everything is loaded in the browser), it guarantees the presence of necessary DOM structure and thus the correct execution of program. Sort of an informal way for doing `$.ready(...)`.

alex-deaconu: @settinghead okay makes sense. Thanks.

src-js/Viewer/Widget/PopupPanel.js
```diff
@@ -0,0 +1,69 @@
+class PopupPanel {
+  constructor() {
+    this.$elem = $('#notificationDiv');
+    this._init();
+  }
+
+  _init() {
+    this.$elem.html('');
+
+    this.$elem.append('<div id="notificationBackdrop"></div>');
```
settinghead: @alex-deaconu I did this mainly because JS code was easier to hot-reload in the dev environment, and also for component encapsulation. Updated as suggested.

src-js/Viewer/Data/ViewerDataController.js
```diff
@@ -131,9 +134,9 @@ var _showDisplayRegistration = function(notificationType) {
 var _showNotification = function(notification) {
   _setClosedState();
 
-  // TODO
+  //TODO: enable this condition
   // if (!service.hasData()) {
```
settinghead: @alex-deaconu I'm still confused about the logic. If I enable this line, seems that notification will never get a chance to show, because `service.hasData()` is always true as soon as you call `reportDataReady()`, which checks and shows notifications. Thoughts?

settinghead:
@alex-deaconu @ezequielc the code has now been updated (in new commit). However, since I enabled some condition checking, I can't seem to create condition that actually shows the popups (see comment above). Please let me know if you have any thoughts. Thanks!

src-js/Viewer/Data/ViewerDataController.js
alex-deaconu: @settinghead hmm. Thought about this some more and there is some confusion here. The issue is not with your code, but with `hasData()`. It should not check for the mere existence of data, it should instead verify if the Viewer has **content**. 

Basically the idea is that the notification should not show if the Viewer already has content playing on the Display, so as to not interrupt the experience. However, on startup the Viewer would get an error code and have no content, so the notification shows.

So, my suggestion is you modify `hasData()` to check for `viewerData.content`. 

Thanks!

src-js/Viewer/Data/ViewerDataController.js
```diff
@@ -79,35 +85,27 @@ var _setNewViewerData = function(newViewerData) {
...
-    break;
-  case Global.VIEWER_STATUS.NO_COOKIE:
-    // _showNotification(NotificationType.cookies_disabled.getMessage());
-    // break;
-
-  case Global.VIEWER_STATUS.CONTENT_NOT_FOUND:
-    // _showNotification(newViewerData.getStatusMessage());
-    // 
-    // break;
-  case Global.VIEWER_STATUS.ID_SHARING_VIOLATION:
```
alex-deaconu: @settinghead as per my previous comment, the `ID_SHARING_VIOLATION` & `DISPLAY_NOT_FOUND` messages are now removed. They **need** to be added back in so we can deal with `_setClosedState()` && show the notification when the Player is not present. See comment below. Thanks!

src-js/Viewer/Data/ViewerDataController.js
```diff
@@ -124,17 +122,16 @@ var _showDisplayRegistration = function(notificationType) {
   //   DisplayRegisterWidget.getInstance().show(notificationType);
   // }
   // else {
-  //   ViewerNotificationsPanel.getInstance().show(notificationType.getMessage());
+  //   ViewerNotificationsPanel.show(notificationType.getMessage());
```
alex-deaconu: @settinghead this code needs to be re-enabled as per the comment above. Thanks!

settinghead:
@alex-deaconu @ezequielc I believe I have fixed everything (3rd commit). Please let me know if I missed something. Thanks!

src-js/Viewer/Data/ViewerDataProvider.js
```diff
@@ -58,7 +58,7 @@ var _dataTimerExecute = function() {
   if (state === WAITING_STATE) {
     // TODO
     // if (!ViewerDataController.hasData()) {
-    //   ViewerNotificationsPanel.getInstance().show(NotificationType.server_connection_failed);
+    //   ViewerNotificationsPanel.show(NotificationType.server_connection_failed);
```
alex-deaconu: @settinghead this use case needs to be re-enabled as well. Thanks.

alex-deaconu:
@settinghead 2 comments in the code above. Also, as mentioned before can you remove the GWT Notification related files:
ViewerNotificationsPanel.java, MessageCounterLabel.java & ViewerNotificationsWidget.java
if they are no longer used?

Thanks!

settinghead:
@alex-deaconu @ezequielc Fixed re comments above. 

I managed to get rid of one class `ViewerNotificationsPanel`. However `MessageCounterLabel` and `ViewerNotificationsWidget` are still being used by some other Java classes, so I'm leaving them in the code for now.

Also, when I run the script locally, I wasn't able to see a display registration popup when no ID or a wrong ID is specified. Is it expected behavior?

alex-deaconu:
@settinghead you can try setting `&player=true` in the URL params. Thanks.

settinghead:
@alex-deaconu OK I see it now. Thanks. PR should be good for another review.

settinghead:
@alex-deaconu FYI I believe `feature/player-controller` also needs a new rebasing to get rid of some of the duplicate commits in this PR.

src-js/Viewer/Data/ViewerDataController.js
alex-deaconu: @settinghead I believe you missed re-adding these two usecases in the JS code. However, It's being dealt with correctly in GWT. Thanks.

alex-deaconu:
@settinghead looks good, thanks for the fixes & rebase. One last comment re- missing use cases. Thanks.

settinghead:
@alex-deaconu done with the missing case.

alex-deaconu:
@settinghead code looks good. However, I'm not seeing any notifications when testing with the deployed version:
http://error-message-dot-rvaviewer-test.appspot.com/?type=display&id=90947cd

It looks like the `notificationDiv` is empty. Am I missing something?

Thanks.

settinghead:
@alex-deaconu currently it only works if you append `&player=true`. Is this not the correct behavior?

alex-deaconu:
@settinghead to clarify, when I enter `player=true` it shows the registration widget. But what I was saying that the NotificationWidget does NOT show for me at all. 

For example:
http://error-message-dot-rvaviewer-test.appspot.com/?type=presentation&id=90947cd
should show the notification of `Presentation not found` similar to:
http://rvaviewer-test.appspot.com/?type=presentation&id=90947cd

Does that clear things up? Thanks.

settinghead:
@alex-deaconu It was a mistake on my part. Should have checked the notification popup is working after code changes. Should be fixed now. Thanks.

src-js/Viewer/Data/ViewerDataProvider.js
```diff
@@ -56,10 +56,9 @@ var _getData = function(url) {
 
 var _dataTimerExecute = function() {
   if (state === WAITING_STATE) {
-    // TODO
-    // if (!ViewerDataController.hasData()) {
-    //   ViewerNotificationsPanel.getInstance().show(NotificationType.server_connection_failed);
-    // }
+    if (!ViewerDataController.hasData()) {
```
alex-deaconu: @settinghead when testing this, there's an error that `Uncaught ReferenceError: ViewerDataController is not defined`. Can you please update with the correct reference? Thanks.

alex-deaconu:
@settinghead also from testing, I'm not seeing messages disappear according to a countdown timer? 

ie. some messages have a countdown (visible or not) that should have them expire after the interval. I don't see support for that in the new code. 

Thanks.

src-js/Viewer/Data/ViewerDataProvider.js
settinghead: @alex-deaconu Sorry again. Fixed.

settinghead:
@alex-deaconu I totally missed that part. Will implement.

settinghead:
@alex-deaconu the counter has been implemented and tested. Please check if there are any other problems. Thanks!

src-js/Viewer/Data/ViewerDataProvider.js
alex-deaconu: @settinghead now I'm getting `Uncaught ReferenceError: ViewerNotificationsPanel is not defined` at the line below. Thanks.

alex-deaconu:
@ezequielc clicking outside of the notification should hide it and interrupt the timer: http://screencast.com/t/PzY1Iy2UG.

Also, one note in the code re: missing reference. Thanks.

ezequielc:
@alex-deaconu I think you meant @settinghead above. Cheers

alex-deaconu:
@ezequielc oops. Sorry about that.

settinghead:
@alex-deaconu reference errors are fixed. Also hiding timed popup should work properly now.

alex-deaconu:
@settinghead from what I can tell everything is working correctly now. Wondering if you can squash, otherwise LG 👍 . Thanks!

PR #131, opened 2016-09-28
Convert ViewerPreviewHeader to JavaScript

settinghead:
new_viewer_preview_header_v2
new_viewer_preview_header_collapsed


settinghead:
@alex-deaconu @ezequielc please review.

Notes on some decisions and trade-offs:

### PopupPanel
- I decided to keep the original class inheritance hierarchy in GWT (i.e. `PopupPanel` -> `NotificationPanel` and `PopupPanel` -> `ViewerPreviewWidget`), because this can be a reusable pattern for two more components (`DisplayRegister` and `HelpFrameWidget`) and was easier to reason about and less error-prone (as opposed to initializing from scratch).

### HTML vs. string templates vs. jQuery dom manipulation
- Because `PopupPanel` remains as a component, the styling of the popup structure is kept reusable. So to avoid code repetition, I moved back some of the stylings back to Popup.js, and used `jQuery.prop()` to initialize them
- For easier editing, I put some of the buttons into partial templates (buttons & panel)
- Some of the stylings are not embedded in HTML/templates but instead is done programmatically via 
`jQuery.css()`. This is mainly because these styles are dynamic and may change as the state of the component changes

### ESLint autofix/prettify has been partially applied to some of the files

- Most files I touched have been automatically fixed/prettfied by my Atom editor
- Note that eslint fixes not only styles, but in some cases it changes the syntax of the files, which should not result in any changes in execution behaviours.

Thanks!

src-js/Viewer/UI/PopupPanel.js
```diff
@@ -1,27 +1,64 @@
...
+  }
 
-    var _this = this;
-    _this.$backdrop.mousedown(() => {
-      _this.hide();
-    })
+  _insertBackdropDiv() {
+    this.$backdrop = $(`
+      <div
+        style="position: absolute; top: 0; bottom: 0; left: 0; right: 0; z-index: 998; background-color: black; opacity: 0.2; ">
```
alex-deaconu: @settinghead wondering if you can take all these inline styles and add them to the `viewer.css` file? Thanks.

src/main/webapp/Viewer.html
```diff
@@ -85,12 +85,8 @@
...
 	</div>
 
-  <div id="notificationDiv" style="position: fixed; top: 0; bottom: 0; left: 0; right: 0; z-index: 997; display: none; ">
-    <div id="notificationBackdrop" style="position: absolute; top: 0; bottom: 0; left: 0; right: 0; z-index: 998; background-color: black; opacity: 0.2; ">
-    </div>
-    <div id="notificationContent" style="position: absolute; padding: 10px; top: 0; bottom: 0; left: 0; right: 0; margin: auto; border: 1px solid #ddd; border-radius: 5px; overflow: visible; font-size: 1.2em; width: 500px; height: 200px; text-align: center; vertical-align: middle; z-index: 999; background-color: white; ">
-    </div>
-  </div>
+  <div id="notification-div"></div>
+  <div id="viewer-header-div"></div>
```
alex-deaconu: @settinghead I'm still not sure why this particular implementation is not:
- Putting all the HTML code for notifications & header in the Viewer.html file
- Populating specific `<div>` or `<span>` text content based on an id/class selector
- Setting `href` properties for links based on an id/class selector
- Adding click handlers for event tracking
- Adding CSS to the `viewer.css` file

In my opinion, the complexity added by creating all these elements dynamically (when they are static elements that don't change) and the technical skill required to make minor changes to styling is not worth the effort.

Am I crazy to say the above? Are there any technical obstacles or reasons why the above is not implementable?

@ezequielc thoughts?

Thanks!

settinghead: @alex-deaconu I think what you are suggesting makes sense. I was thinking strictly from a functional component's point of view (i.e. encapsulate all popup logic and functional styling in one single place in `PopupPanel.js` like a React/Angular component), but I agree with only jQuery, it adds to code complexity. I'll change according to your suggestions and see how it goes.

settinghead:
@alex-deaconu got rid of most of jquery populating code based on your suggestion. Please review. Thanks!

alex-deaconu:
@settinghead looks good. Thanks for the refactor.

Couple of things:
Header is not showing up here: http://preview-header-dot-rvaviewer-test.appspot.com/?type=presentation&id=c219885a-f1c9-4d03-94c5-cf1ad9353e87
(Note, this presentation is a template, that may have something to do with it).

Also, there are some styling differences between the old and new headers. Wondering if you can correct as best you can, I don't think we need to be 100% for this.

Thanks.

settinghead:
new_viewer_preview_header_v2
new_viewer_preview_header_collapsed


@alex-deaconu @ezequielc  Fixed the issue. It was due to a missing method, but there were no errors on the console probably because it was handled by try/catch.

I also made an adjustment to the styling (drop shadow). Let me know if there is anything else in particular that you think needs a fix. Thanks.

alex-deaconu:
@settinghead I believe both the `top` and `left` CSS should be `-10px`. 

Here's a comparison between existing and this: http://screencast.com/t/XiApgKKm, I'm not sure if the remaining styling differences are worth pursuing. 

@ezequielc thoughts? Thanks.

alex-deaconu:
@settinghead looks good now. Waiting to hear from @ezequielc re: styling. Thanks.

ezequielc:
@settinghead The header is a bit taller than previous version. If you can set its padding to 0px, it will make it about the same size. I am good with everything else. Thanks!

settinghead:
@ezequielc padding removed as suggested. I will squash & merge if everything looks fine. Thanks.

ezequielc:
@settinghead Looks good to me. Thanks!

ezequielc:
@settinghead DO not merge yet

ezequielc:
@settinghead @alex-deaconu We need to make sure we will not overwrite the production version if we merge here. Also, as mentioned in Delivery Room, we need to hold deploying Viewer until https://github.com/Rise-Vision/rise-launcher-electron/issues/483 is resolved. Thanks

PR #132, opened 2016-09-29
Fix issue playing MOV files with Rise Cache v2

ezequielc:
@alex-deaconu I need to deploy a version in production environment with a fix for playing MOV files in RCv2. Alan is going to verify this and perform the QA. 
Would it be ok to deploy as configured here until we work on our improvement card?
Thanks


alex-deaconu:
@ezequielc looks good 👍 . Wondering if you could comment out the line with the old code instead of removing it; I'm thinking it's easier to revert. Thanks.

ezequielc:
@alex-deaconu Updated. Cheers!

circle.yml
```diff
@@ -30,10 +30,11 @@ deployment:
     branch: /(feature|fix|chore).*/
     commands:
       - ./node_modules/.bin/cross-env NODE_ENV=dev npm run build
-      - mvn appengine:update -Dappengine.version=$(echo -n $CIRCLE_BRANCH |awk 'BEGIN{FS="/"}{print tolower($NF)}') -Dappengine.appId=rvaviewer-test
+      - mvn appengine:update -Dappengine.version=qa -Dappengine.appId=rvaviewer-test
```
alex-deaconu: @ezequielc is it necessary to stage to `qa` version on Stage too? I don't think we need to change this... Thoughts? Thanks.

alex-deaconu:
@ezequielc looks good 👍 . Thanks for fixing!

PR #135, opened 2016-09-30
Content rendering

alex-deaconu:
WIP Branch for rendering and cycling functionality.


package.json
```diff
@@ -38,6 +38,7 @@
     "gulp-uglify": "^2.0.0",
     "gulp-util": "^3.0.7",
     "jquery": "^3.1.0",
+    "underscore": "^1.0.0",
```
ezequielc: @alex-deaconu I don't think you are using this dependency. Should we remove it? Thanks

src-js/Viewer/Content/PlaceholderController.js
```diff
@@ -0,0 +1,473 @@
...
+      } else if (gadgets.length !== 1) {
+        _playNextItem(executeDoneCommand);
+      } else {
+        currentItem = nextItem;
+      }
+    } else if (!gadgets[nextItem].isReady()) {
+      if (gadgets.length !== 1){
+  //				playNextItem();
+        
+        var duration = gadgets[nextItem].getItem().duration;
```
ezequielc: @alex-deaconu Should we have a default value like in the original in case duration is missing: `RiseUtils.strToInt(gadgets.get(currentItem).getItem().getDuration(), 0);`? Thanks

src/main/webapp/Viewer.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
 
-<html  manifest="offline.manifest" itemscope itemtype="http://schema.org/Product" prefix="og: http://ogp.me/ns# fb: http://ogp.me/ns/fb#">
```
ezequielc: @alex-deaconu Curious, why did you remove `manifest`? Thanks!

package.json
alex-deaconu: @ezequielc good question. Actually it's supposed to be used in the `PresentationParser`, however for some reason it works without it. I will double check if it's needed or not. Thanks.

src-js/Viewer/Content/PlaceholderController.js
alex-deaconu: @ezequielc agreed, that would be a good idea.

src/main/webapp/Viewer.html
alex-deaconu: @ezequielc actually this and the gulpfile changes are part of my own development environment changes. It helps me reload the page without using Incognito (and speed up the build/test iterations). I will revert both files for a final PR; please ignore. Thanks.

alex-deaconu:
@ezequielc my latest commit implements a simplified Video Item in the Viewer. To summarize:

The old Video Item had a couple of features which were implemented for certain reasons and as certain workarounds for bugs with Videos. One feature is a *nudge* mechanism, which detects if a Video has stopped playing and is no longer responding. In these cases it will try to '*forward*' the video 3 times, and if that fails it will just `window.reload()` the Viewer. The second feature is a Video controller, which loads at most 3 videos per placeholder at one time (previous, current and next). This is obviously a solution for placeholders with 100s of videos playing where loading all the Videos into memory would cause issues.

Not sure if those issues are relevant or in my opinion if we care that the issues occur anymore. I'm saying this because we are trying to push users away from the Video item in the first place (to use the Video Widget). So, I decided to simplify the Video Item implementation to just render the Video and play it. Both those 2 features, removed.

Wondering if you can take a look and let me know your opinion on the above. Thanks!

ezequielc:
@alex-deaconu  My concern is that by removing them we can decrease the reliability of the Viewer. How hard would it be to implement these features? Thanks!

alex-deaconu:
@ezequielc the issue is not necessarily complexity of implementation. The features themselves are prone to errors and very tough to verify. Will implement as per your suggestion. Thanks.

ezequielc:
@alex-deaconu I see your point. As we are migrating existing logic, as long as we keep the current behaviour, I think we are good. 
Also, I finished the initial review. It looks good to me. We can do a more detailed review when adding the unit tests. Thanks!

alex-deaconu:
@ezequielc updated PR with the complete Video Item implementation from GWT, as per discussion. Thanks.

alex-deaconu:
Changing PR source to `feature/display-registration` branch.

PR #145, opened 2016-10-18
Add multi-schedule test

tejohnso:
@ezequielc @fjvallarino please review


fjvallarino:
:+1:

ezequielc:
@tejohnso 👍 
Quick question, do you think it would be complicated/unreliable to check for the duration of each Presentation as well? Thanks!

tejohnso:
@ezequielc I think it's reasonable for another card and complexity would depends on the expected timing precision.

PR #146, opened 2016-10-18
Re-enable Data retrieval from V3 player

alex-deaconu:
Move feature to JS
Hide display registration dialog


alex-deaconu:
@ezequielc @tejohnso this is my suggested implementation for the Data Retrieval from the Player (in Javascript). These changes are also integrated into `feature/rendering` branch. Can you please take a look and let me know your thoughts? Thanks.

src-js/Viewer/Data/ViewerDataProvider.js
```diff
@@ -76,8 +76,6 @@ var _dataTimerExecute = function() {
 };
 
 service.retrieveData = function(reason) {
-  // if (window.disableViewerContentFetch) {return;}
-
```
tejohnso: @alex-deaconu I think we need to make sure it's disabled here.  There are other calls into retrieveData.  From what I've found, channel content change and polling timer events both use this and we want both to be disabled.  If we don't keep the flag awareness here, then we'll have to go the other route of disabling them at call site.

alex-deaconu: @tejohnso good points. 

Re Channel, it is only initialized here:
https://github.com/Rise-Vision/viewer/pull/146/files#diff-b64c8635eef626e3a3f880985b19557aL69
which will not happen as this is now skipped due to the refactoring w/ new `_updateViewerData` function.

However, the polling timer issue is a valid one. See second commit. I've moved the polling initialization and messages being shown to not occur when the data is retrieved from the v3 player. Also, I've re-added the check for the `window. disableViewerContentFetch` flag. 

Thanks!

tejohnso:
@alex-deaconu Do you think we can also add the flag check in RisePlayerController's callUrlNative?  We shouldn't bother trying to contact java player if the flag is set.

alex-deaconu:
@tejohnso hmm, doesn't the `v3` player expect the Viewer to send a heartbeat every 1 minute? I'm assuming it does.

Also, as a test was failing, I've overwritten the commit from yesterday with a new one. This one removes yesterday's change to the DataController (as there was an issue) and adds tests. Please review.

Thanks.

tejohnso:
@alex-deaconu Any `v3` players with the flag set will be responding via ipc as [seen here](https://github.com/Rise-Vision/viewer/blob/feature/rendering/src-js/Viewer/Player/RisePlayerController.js#L142)

alex-deaconu:
@fjvallarino @settinghead FYI this is the PR for retrieving data from v3 player. Thanks!

PR #151, opened 2016-10-20
Pick 0 index when Id is not matched

alex-deaconu:
Fixed logic to be similar to GWT implementation:
https://github.com/Rise-Vision/viewer/blob/1201c697d6f861fceba07f4f400aa755983a6791/src/main/java/com/risevision/viewer/client/data/ViewerDataControllerBase.java#L88

Basically, if the Presentation is not found, use the 0 index. 

Fixes issue with Demo content.

@ezequielc @settinghead @fjvallarino please review. Thanks!


fjvallarino:
:+1:

settinghead:
@alex-deaconu :+1:

PR #155, opened 2016-10-24
Fix issue where header blocks the entire page

settinghead:
Card: https://trello.com/c/RqWU65fI/1762-issue-148-interactions-don-t-work-in-preview-2


settinghead:
@ezequielc @tejohnso @fjvallarino the issue should be now fixed and the content should be interactive when a header is present. To verify, please check this URL:
http://cannot-interact-with-header-dot-rvaviewer-test.appspot.com/?type=presentation&id=d64d0c04-60a5-4ab0-94ff-cc1fd3cd97d8

Please review. Thanks!

ezequielc:
@settinghead 👍 Thanks for fixing it!

PR #158, opened 2016-10-26
Use function keyword for PlaceholderController

tejohnso:
Arrow functions cannot be used as constructors

@settinghead @ezequielc please review.  This came up as an issue when I ran browserify directly without es2015 conversion.


settinghead:
@tejohnso :+1:

PR #170, opened 2016-11-07
handle case where item.recurrenceDaysOfWeek is undefined

settinghead:


settinghead:
@ezequielc @rodrigopavezi please review here: 
https://github.com/Rise-Vision/viewer/pull/170/files?w=0

tejohnso:
@settinghead a few suggestions:
 - Whitespace corrections in a separate commit
 - Use str.includes() rather than str.indexOf() !== -1
 - Replace the seven line conditional with a function call

settinghead:
@tejohnso 

- I turned off auto-linting feature in my editor and decided to do all the linting fixes at another time. It's simply distracting for the reviewers.
- I tried `str.include()` but failed some unit tests because PhantomJS doesn't seem to support it. So I had to stick with `str.indexOf() !== -1`
- Created a new function to check recurrence day conditions per your suggestion.

tejohnso:
:+1:

PR #171, opened 2016-11-07
Fix/channel controller timing

tejohnso:
@alex-deaconu Can you clarify why that `state !== INACTIVE_STATE` check was there?


alex-deaconu:
@tejohnso `state !== INACTIVE_STATE` just makes sure the Controller was initialized. This also includes the check for a Display, and that it is not an Embedded Presentation. While that state is necessary for not "double initializing" the controller, I'm not sure if it is needed in `setChannelMessage` (ie. a Message **should** not be received if the controller is inactive).

Thanks.

src-js/Viewer/Channel/ChannelConnectionController.js
```diff
@@ -67,39 +67,36 @@ service.setChannelToken = (channelToken) => {
...
-      connectionVerificationTimer = setTimeout(_connectionVerificationTimerExecute,
-        (ViewerDataInfo.getPingInterval() + 3) * global.MINUTE_UPDATE_INTERVAL);
-    } else if (message.indexOf(channelConnectionProvider.MESSAGE.UPDATED) !== -1) {
-      updateTicket = message.substring(channelConnectionProvider.MESSAGE.UPDATED.length).trim();
+    channelConnectionProvider.channelConnected();
 
-      RiseVision.Viewer.Data.ViewerDataProvider.retrieveData(RiseVision.Viewer.Data.ViewerDataProvider.REASON.UPDATE_MESSAGE_RECEIVED);
+    _connectionVerificationTimerClear();
+    connectionVerificationTimer = setTimeout(_connectionVerificationTimerExecute,
+    (ViewerDataInfo.getPingInterval() * 4) * global.MINUTE_UPDATE_INTERVAL);
```
alex-deaconu: @tejohnso as per our discussion this morning, I mentioned that if a `PING` is not received within the interval, the Viewer would revert to Polling. However, with this implementation the interval is `4 * PING_INTERVAL` so only after an hour of not receiving PING, the Viewer will switch to Polling, and Poll after the `POLL_INTERVAL` timeout (ie. 1h 15m later).

Having said that, I think we need to re-think this logic; the first Poll comes too late in this scenario. 

Thoughts? Thanks.

tejohnso: @alex-deaconu I think that's okay.  We wait an hour for a ping.  If we don't get it, we start polling.  The first poll has that extra 15mins but I don't think that's a big deal.

alex-deaconu: @tejohnso okay. I don't disagree. However, we should double check with Alan tomorrow in SoS as this was not the logic he signed off on. Thanks.

ezequielc: @alex-deaconu I tend to agree with @tejohnso but it is a good point to review with Alan. Thanks!

tejohnso:
@alex-deaconu I left the ACTIVE_STATE checking as is in master.  But I included a refactoring of the conditionals.  Please have a look.  Note that there is still a situation where ACTIVE_STATE is being set on a MESSAGE.CONNECT.

src-js/Viewer/Channel/ChannelConnectionController.js
```diff
@@ -67,39 +67,40 @@ service.setChannelToken = (channelToken) => {
...
+  }
 
-      _connectionVerificationTimerClear();
-      connectionVerificationTimer = setTimeout(_connectionVerificationTimerExecute,
-        (ViewerDataInfo.getPingInterval() * 4) * global.MINUTE_UPDATE_INTERVAL);
-    } else if (message.indexOf(channelConnectionProvider.MESSAGE.UPDATED) !== -1) {
-      updateTicket = message.substring(channelConnectionProvider.MESSAGE.UPDATED.length).trim();
+  if (message === channelConnectionProvider.MESSAGE.PING ||
+  message.indexOf(channelConnectionProvider.MESSAGE.UPDATED) !== -1 ||
+  (message === channelConnectionProvider.MESSAGE.CONNECT && state !== INACTIVE_STATE)) {
```
alex-deaconu: @tejohnso I think this use case is redundant. As we only have 3 states (`INITIAL`, `INACTIVE` & `ACTIVE`), and we reject `INITIAL_STATE`, if `state !== INACTIVE_STATE` then it is already `ACTIVE_STATE`. Do you think we should remove it? Thanks.

tejohnso:
@alex-deaconu That scenario was made more clear after refactoring and I was going to bring it up.  I've removed it in 497eff3.

alex-deaconu:
@tejohnso thanks for fixing. Looks good 👍 . Thanks.

PR #175, opened 2016-11-08
Set load verification time for 1 minute

alex-deaconu:
As Presentations are loaded and unloaded as needed,
a Preload buffer was added; previously set to 10 mins

However, if an item is set to play for less than 10 minutes,
it would not be loaded and hence wouldn't play

Changed the buffer time to 1 minute to aleviate the issue,
however if an item is set to play for only 1 minute,
it will not load & play.


alex-deaconu:
Potential fix for issue #172. This is a quick fix; explanation re: caveat above. 

settinghead:
@alex-deaconu :+1:

PR #182, opened 2016-11-15
Integration tests

settinghead:
- Schedule/presentation/placeholder/playlist


settinghead:
@alex-deaconu @ezequielc @fjvallarino I had to single out integration tests from unit tests, as some of the other unit tests was interfering with the setup (the integration tests would fail if I mix them together).

Also please note that I this PR does not include the scenario where a play item is set to done and is restarted.

Please review and let me know if the approach in `karma.confg.js` is OK. Thanks!

alex-deaconu:
@settinghead do you think we have an overarching problem that is causing the test suite to fail when running all tests? This is not the first time you mention this occurring. Thanks.

settinghead:
@alex-deaconu yes, I do think so. The root cause is environment is not completely isolated for each unit test, and changes to globals would leak to other places.

Also I think maybe JSDom could help us with this issue, because it provides completely isolated (but simulated) browser environment for each test. I tried it before and it worked great for some early unit tests we wrote.

settinghead:
@alex-deaconu @ezequielc @fjvallarino I restored `karma.config.js` and also marked all tests in this PR as skip. If this looks OK I'll merge it, and fix the unit test interference issue in the next PR. Thanks!

package.json
```diff
@@ -5,7 +5,9 @@
   "main": "src-js/index.js",
   "scripts": {
     "test": "karma start ./karma.config.js",
-    "test:ci": "karma start ./karma.config.js --single-run && gulp test:e2e",
+    "test:unit": "karma start ./karma.config.js --single-run",
+    "test:integration": "TEST_SET=integration karma start ./karma.config.js --single-run",
+    "test:ci": "npm run test:integration && npm run test:unit && gulp test:e2e",
```
alex-deaconu: @settinghead I believe you missed reverting these changes. Thanks.

settinghead: @alex-deaconu Good catch. Updated.

settinghead:
@alex-deaconu renamed the file to `Content.integration.tests.js`. Thanks.

alex-deaconu:
@settinghead looks good, thanks for fixing 👍

PR #187, opened 2016-11-17
Exception logging into BQ

settinghead:


settinghead:
@ezequielc @alex-deaconu @tejohnso please review via this link: https://github.com/Rise-Vision/viewer/pull/187/files?w=0.

Note: I did not save exception into a separate table and instead used a `Exception - ` prefix and saved everything into the `Viewer_Event` table. Let me know if this looks right. Thanks.

src-js/Viewer/Logger.js
```diff
@@ -12,6 +12,24 @@ Logger.logExternalMessage = (eventName, eventDetails) => {
   window.logExternal(eventName, window.RiseVision.Viewer.EntryPoint.getDisplayId(), Global.VIEWER_VERSION, eventDetails);
 };
 
+Logger.logException = (methodName, err, identifier) => {
+  try {
```
alex-deaconu: @settinghead wondering why you chose to add a new function and modify all those files rather than just add BQ logging functionality and improve to the Logger.log function. Thoughts? Thanks.

src/main/webapp/scripts/ext-logger.js
```diff
@@ -95,11 +95,50 @@
       xhr.setRequestHeader("Content-Type", "application/json");
       xhr.setRequestHeader("Authorization", "Bearer " + EXTERNAL_LOGGER_TOKEN);
       xhr.send(JSON.stringify(insertData));
-      
+
       console.log("Log External Event", insertData.rows[0]);
     }
   }
 
+  function logException(eventName, displayId, version, eventDetails) {
```
settinghead: It was mainly a risk-averse strategy so as not to touch existing code in production. I'll take another look at if some of the logic can be safely merged. Thanks.

settinghead:
@alex-deaconu updated code to use `window.logExternal`. 

I had to keep `Logger.logException` which now calls `Logger.logExternalMessage`. The reason for the extra method is that it does a few extra things: append a `Exception -` prefix to the event, accept and incorporate an exception identifier/locator as parameter, and also convert error object to string.

src-js/Viewer/Parser/GadgetSettingsParser.js
```diff
@@ -83,36 +83,37 @@ var GadgetSettingsInfo = function(gadgetUrl, additionalParams, gadgetXml) {
...
+
   var _parseGadgetXml = (gadgetXml) => {
     var d;
     try {
       d = _parseXmlObject(gadgetXml);
-    } catch (e) {
-      Logger.log('Error parsing XML', e);
+    } catch (err) {
+      Logger.log('Error parsing XML', err);
+      Logger.logException('GadgetSettingsParser._parseGadgetXml', errl);
```
tejohnso: @settinghead Is `errl` a typo?

src-js/Viewer/Logger.js
alex-deaconu: @settinghead not sure if you saw this question. My suggestion is to merge the `Logger.log` & `Logger.logException` functions as they practically do the same thing and this way we don't have 2 log functions per catch statement. Thanks.

src-js/Viewer/Parser/GadgetSettingsParser.js
settinghead: @tejohnso definitely yes. Good catch.

tejohnso:
:+1:

src-js/Viewer/Logger.js
alex-deaconu: @settinghead the idea is that instead of having 2 functions per catch statement `log` and `logException`, we only have 1. This function would log to console, viewer log and BQ. This way we can be consistent with Exception logging across the board. Thanks.

settinghead:
@alex-deaconu @tejohnso I made changes to consolidate writeLog, console.log and writing to BigQuery data into one function.

I had to squash some of my commits to reduce overhead for rebasing. Sorry about any confusion caused. Please review. Thanks!

src-js/Viewer/Logger.js
```diff
@@ -4,12 +4,41 @@ const Global = require('./Config/Global');
...
+  if (additionalData) {
+    additionalData = JSON.stringify(additionalData);
+  }
+
+  const viewerLogMsg = additionalData ? `${eventName} - ${additionalData}` : eventName;
+
+  window.writeToLog(viewerLogMsg);
+  if (additionalData) {
+    /* eslint-disable no-console */
+    console.log(eventName, eventDetails, additionalData);
```
alex-deaconu: @settinghead I found it really helps with debugging if the Exception object (for the stack trace) was added to the console. Wondering if you can do that instead, similar to how `Logger.log` is above. Thanks.

alex-deaconu:
@settinghead looks great, thanks for all the fixes 👍 !

Issue #190, opened 2016-11-21
If a Widget fails to send "ready", we do not retry


alex-deaconu:
Currently we have a retry mechanism for Gadgets and Image items, but not Widgets.

Reported in the community here: https://community.risevision.com/rise_vision_inc/topics/mystery-of-the-missing-widgets

Decision is to add feature.

alex-deaconu:
@btfrise @joelasso please inform the user that this has been corrected. Wondering if it will rectify the situation... Thanks!

PR #191, opened 2016-11-21
Re-try widget load if ready is not received within 1 min

alex-deaconu:
Log failures to BQ


alex-deaconu:
Fixes #190 

Functionality similar to Gadget & Image Item retry mechanisms from the same file.

To verify, create a Presentation with an invalid Widget URL (ie. modify the HTML to break the URL). Run that Presentation in the Viewer. You should see in the Viewer Log Div a retry counter, and 3 retries (one every minute) and a failure message. 

fjvallarino:
:+1:

settinghead:
@alex-deaconu :+1:

PR #192, opened 2016-11-21
Improve linting (applies no-var rule and prefer-const)

fjvallarino:
The last commit was added for handling a situation were a const variable was suggested but, even if it was correct since it was never modified, it did not make sense based on the variable name. 


tejohnso:
:+1:

alex-deaconu:
👍

PR #193, opened 2016-11-21
Set indent rule to two spaces

fjvallarino:
I was wondering if applying this rule was worth it, since it affects history (it makes it harder to compare to older commits), but it's not as bad as it could have been. Let me know if you'd rather not apply it.


tejohnso:
@fjvallarino in my opinion it's worth it for the consistency going forward.

alex-deaconu:
@fjvallarino 👍 . Agreed with @tejohnso 's comment. Thanks.

PR #195, opened 2016-11-21
Standardize quotes

fjvallarino:
This one is controversial. The "natural" option was choosing single quotes, since it's what was being used most in the project (by far), which means a lot less files needed to be changed. I chose double quotes because it's the suggested standard and, mainly, because single quotes required a lot of character escaping. In case you disagree, I'll switch to single quotes.


alex-deaconu:
@fjvallarino my preference is single quotes, but I'm curious what everyone else thinks. Thanks.

tejohnso:
@fjvallarino I agree with the suggested standard of using double quotes.  It's a good habit for the easier escaping and also because it's the JSON standard so you have no choice but to use double quotes at least some times.

settinghead:
@fjvallarino I'd vote for double quote. A previous practical reason for using single quote is for easier coding for inline HTML strings (as HTML attribute values use double quotes). However, this benefit is no longer as relevant because we can now use ES6 String template literals for HTML strings.

fjvallarino:
@Rise-Vision/apps This is currently at 3 votes for double quotes, 1 for single. I'm going to merge for the moment (it blocks me on the rest of the rules), and we can review later. Thanks all for voting!

alex-deaconu:
@fjvallarino double quotes works too (: . 👍

Issue #189, opened 2016-11-18
White Screen when for multiple instances of same Presentation in Schedule


ezequielc:
I could not narrow down the issue, but I am opening the issue for investigation.

When adding the same Presentation multiple times in a Schedule, set to weekly and each item using different weekdays, a white screen is showing in the display.

Sample Presentation: https://apps.risevision.com/schedules/details/0a5e3d78-5102-4c7f-9d2f-8b4541ad991d?cid=f114ad26-949d-44b4-87e9-8528afc76ce4
Sample Display: http://preview.risevision.com/?type=display&player=true&id=JUGACBYRY8Z9

alex-deaconu:
@ezequielc this issue has been fixed by the above commit. I've validated the fix by setting my clock to Friday as suggested. 

Wondering if you could verify & update the User this issue is relevant to.

Thanks!

ezequielc:
@alex-deaconu Thanks! I validated and notified the user. Cheers!

alex-deaconu:
@ezequielc thanks!

PR #196, opened 2016-11-21
Update variable reference

alex-deaconu:
`ScheduleController` is not available as a variable
in the `PresentationController` due to circular reference


alex-deaconu:
Fix for #189

fjvallarino:
:+1:

PR #198, opened 2016-11-22
disable logging for exceptions that are known issues

settinghead:


settinghead:
@tejohnso @alex-deaconu please review. Thanks.

alex-deaconu:
@settinghead I don't think commenting out the code is a proper solution for this. Can you revert back to regular logging via `Logger.log` for those particular use cases? Thanks!

src-js/Viewer/Utils.js
```diff
@@ -178,7 +178,7 @@ utils.reportViewerConfigToPlayer = () => {
   try {
     window.postToPlayer({ message: "viewer-config", viewerVersion, width, height });
```
alex-deaconu: @settinghead wondering if it would be better to check for the function's existence and log if it doesn't exist, rather than use `try/catch` here. Thanks.

src-js/Viewer/Player/RisePlayerController.js
```diff
@@ -115,7 +115,7 @@ service.setPlayerInfo = function (data) {
         Logger.logExternalMessage("player command", command);
       }
     } catch (err) {
-      Logger.logException("RisePlayerController.setPlayerInfo", err);
+      // Logger.logException("RisePlayerController.setPlayerInfo", err);
```
alex-deaconu: @settinghead I don't believe we had issues with this line. It was `_pingV3Watchdog` that was the top log entry; wondering if you can revert this and update that one. Thanks.

settinghead:
@alex-deaconu updated.

src-js/Viewer/Player/RisePlayerController.js
```diff
@@ -151,7 +151,7 @@ let _pingV3Watchdog = function () {
       from: "viewer",
     });
   } catch (err) {
-    Logger.logException("RisePlayerController._pingV3Watchdog", err);
+    Logger.log("Exception - RisePlayerController._pingV3Watchdog", err);
```
alex-deaconu: @settinghead wondering if you can do the same check for `window.postToPlayer` existence here as well. Thanks.

alex-deaconu:
@settinghead one additional suggestion. Otherwise looks good. Thanks!

settinghead:
@alex-deaconu updated.

I realized it might be too risky to remove try-catch block even after we are checking for the existence of `window.postToPlayer`, because other types of exceptions might arise from unexpected places. So I'm adding try-catch back to the two cases mentioned above, in addition to existence checking. Thanks.

src-js/Viewer/Player/RisePlayerController.js
```diff
@@ -145,13 +145,17 @@ let _pingPlayer = function () {
...
 let _pingV3Watchdog = function () {
-  try {
-    window.postToPlayer({
-      message: "ping",
-      from: "viewer",
-    });
-  } catch (err) {
-    Logger.logException("RisePlayerController._pingV3Watchdog", err);
+  if (!window.postToPlayer) {
+    Logger.log("WARN - RisePlayerController._pingV3Watchdog: window.postToPlayer does not exist.", err);
```
alex-deaconu: @settinghead I believe `err` is undefined here. Thanks.

alex-deaconu:
@settinghead I noticed that, good thinking! One typo, otherwise looks good. Thanks.

alex-deaconu:
@settinghead thanks for fixing. 👍

PR #201, opened 2016-11-25
Asynchronous request for the Video URL

alex-deaconu:
Rise Cache URL request moved to the Video Item


alex-deaconu:
This fixes the issue with mistiming between Rise Cache response & Video Item initialization.

The main problem is that the *first* Video Item is initialized when the Viewer API response is received. So, to fix that we can either block all Viewer initialization until Rise Cache is initialized (around 500ms), which didn't seem like a good solution considering the Video Item is deprecated.

So, I moved responsibility for getting the Video URL via Rise Cache to the Video Item itself. The `parent.parent...` call is ugly, but it works.

alex-deaconu:
This has already been tested by CS. Needs to be released.

@ezequielc @tejohnso can you please review? Thanks!

ezequielc:
@alex-deaconu 👍

PR #205, opened 2016-12-14
Hide old error messages after clearing presentations

fjvallarino:
Can be verified on: http://hide-no-presentations-message-dot-rvaviewer-test.appspot.com/Viewer.html


ezequielc:
@fjvallarino Could you provide a link to a display (Viewer URL) where I could replicate the issue and the fix in the new version? Thanks

fjvallarino:
@ezequielc I don't think you can test this outside Player, since it depends on receiving an update message. For testing you can use Q2MNN6WRHSRC, and the error happens with production Viewer (the one used by Player by default). The steps to reproduce are:

- Remove the presentation from the schedule, click save
- Error message shows with black background
- Re add the presentation to the schedule, click save
- Error message shows over the presentation

You can then add:

viewerurl=http://hide-no-presentations-message-dot-rvaviewer-test.appspot.com/Viewer.html

to RiseDisplayNetworkII.ini, restart Player and execute the steps once again.

alex-deaconu:
@fjvallarino wondering if you can move the `hide()` command here: https://github.com/Rise-Vision/viewer/blob/master/src-js/Viewer/Data/ViewerDataController.js#L114
and remove it from here https://github.com/Rise-Vision/viewer/blob/master/src-js/Viewer/Data/ViewerDataController.js#L120. 

This way all notifications would be cleared when new Data is received.

Thanks!

alex-deaconu:
@fjvallarino sorry again for the confusion re: line numbers. 

Fix looks good to me 👍 (although I did not verify on the player). Thanks!

PR #207, opened 2016-12-19
Only stop timer propagation if an item is changed

alex-deaconu:
Sample presentation to verify: http://stop-propagation-dot-rvaviewer-test.appspot.com/?type=presentation&id=45685a53-a977-4958-8edf-7c1cc5fb2bdb
Placeholder counter should run until 59 seconds.

@fjvallarino please review. Thanks!


fjvallarino:
@alex-deaconu is this if clause still needed? https://github.com/Rise-Vision/viewer/blob/85f0939a38c9dff398fe2d966c4f3375eaceaeed/src-js/Viewer/Content/TimerController.js#L31

fjvallarino:
@alex-deaconu I guess `stopPropagation` can be modified by the `timerCommands[i]();` calls. If that's the case, :+1:

alex-deaconu:
@fjvallarino correct. `stopPropagation` is called within the commands itself to stop calling the rest of the stack. However, in our case it was called all the time and in the case of a single item in the Schedule with a specified duration, it would call `stopPropagation` every x duration seconds. Each time this call was made, a timer beat was skipped causing the longer duration for the item. Thanks.

PR #209, opened 2017-01-10
Double timer interval after every API failure

alex-deaconu:
Reset interval on success
Max interval is pollingInterval


alex-deaconu:

ezequielc:
@alex-deaconu Code looks good. Any suggestion on how I could test this? Thanks

alex-deaconu:
@ezequielc best way is to turn off your wifi, and run a Display URL. It should not be able to get an API response and poll after 1 minute, 2 minutes, etc.

Actually, I just realized there's an issue with the Notification for "Retrying in 60s" and its counter; it is hardcoded to 60 seconds so it doesn't reflect the actual interval. Rather than showing the full timer my suggestion is we only show the notification: "An error occurred trying to connect to the Server. Retrying..." for 10 seconds on retry. It will simplify the code and be less distracting for anyone looking at the Display. Thoughts? 

Thanks.

ezequielc:
@alex-deaconu Ok, I will try that. 
Re. Notification, if you can't use the actual number, I agree with the proposed solution. Thanks

alex-deaconu:
@ezequielc well, currently the number is hardcoded, so the change is not major. However, I don't think it would be too exciting for the User to see a countdown of 900 seconds (15 minutes). Maybe we can just show "Retrying in 1 minute..." (2 minutes, etc) without the countdown. Thanks.

ezequielc:
@alex-deaconu Good point. I was wondering if we should just state 'Retrying...' without the minutes/seconds, or maybe use MM:ss format, but in the end, this will not happen too often as the message would only show on initialization, right?
 I think it is ok to show the seconds as the user will see that they increase as it keeps failing. I don't think we need to make this too complicated. Thanks

alex-deaconu:
@Rise-Vision/apps second commit is ready. Just passing the amount of seconds until the next Viewer API call. Thanks.

fjvallarino:
:+1:

ezequielc:
@alex-deaconu Just a heads up, I did some connection on/off tests and left the Viewer running with content but disconnected for around 30 minutes and after that I got an Oh Snap error. It might be because I had devtools opened. I am running it again without devtools to double check, but if you want to leave it running for a while too, so we can test in parallel. Thanks

ezequielc:
@alex-deaconu I think it is something wrong with my Presentation actually, even the production version is leaking memory and leading to Aw, Snap. I will try another Presentation. Thanks

alex-deaconu:
@ezequielc hmm, interesting. Just did some more tests and wasn't able to replicate this behaviour. Curious, what does your Presentation contain? Can you send me the link? Thanks.

ezequielc:
@alex-deaconu My Presentation was https://apps.risevision.com/editor/workspace/5ac770b5-a2e8-4cb6-bc61-6211feffa80b/.  Unfortunately I changed the Schedule, but I had a Schedule playing 3 items of that presentation for 10 seconds each.
I started the Viewer offline, waited for the warning to show and turned connection on. After content was loaded, I turned the connection off again. I can check memory usage increasing overtime.

alex-deaconu:
@ezequielc to confirm, the issue discussed above is fixed with the PR that was just merged. Merging this as well. Thanks!

ezequielc:
@alex-deaconu 👍 Thanks

PR #210, opened 2017-01-17
Fix issue with Rise Cache calls doubling

alex-deaconu:


ezequielc:
@alex-deaconu 👍 Thanks for fixing!

PR #212, opened 2017-01-19
Adjust timers so next schedule/item 60 seconds checks fall at the beginning of the next minute

fjvallarino:
Originally I had modified _setNextScheduleCheck and _setNextItemCheck to handle the 60'' internally, but since _setNextItemCheck is also used for items durations it would have caused issues. Considering this, I kept the same logic for _setNextScheduleCheck be consistent (although in this case it would have been fine).


alex-deaconu:
@fjvallarino looks good 👍 . Thanks!

alex-deaconu:
@fjvallarino actually, sorry, I didn't read your note. I'm not sure why the seconds are subtracted from `_setNextItemCheck`? Also, wouldn't it make sense to add the date subtraction inside the `_setNextScheduleCheck` function? Thanks.

fjvallarino:
@alex-deaconu for _setNextItemCheck the logic I applied is the same as for _setNextScheduleCheck: if you are requesting a delay of 60 seconds, it makes sense to only wait enough for the minute to change (since Timeline checks depend on the minute count changing). Of course, this does not apply if you are setting it to a different value (for instance, duration), which is the reason I moved it out of the function itself as I had done at the beginning.

The case of _setNextScheduleCheck is different since, as far as I see, only two values are used: -1 and 60. Since 60 (the case we're interested in handling differently) is only used twice in PlaceholderController and once in ScheduleController, it was not a big advantage to have it as part of the function itself (that's why I preferred to handle it the same way as _setNextItemCheck, to keep some sort of consistency). I can move it back in, of course.

fjvallarino:
@alex-deaconu when I say "requesting a delay of 60 seconds", I mean explicitly requesting for 60 seconds hard coded in Javascript, since it clearly indicates waiting for "one minute" or long enough for Timeline checks to return something different

alex-deaconu:
@fjvallarino okay, I'm good with leaving the `_setNextScheduleCheck` as is.

However, the `_setNextItemCheck(60)` is just a ready check, which only happens when items are not ready. It doesn't have to be on the minute since it is not associated with a Timeline, so in my opinion the second subtraction is not needed in that case.

Thanks!

alex-deaconu:
@fjvallarino okay, see what you mean regarding the `_setNextItemCheck`, it is associated with a timeline so it does make sense to check in that manner. Disregard the last comment. Thanks.

PR #214, opened 2017-01-19
Removing dependency on isPlaying for the play function

alex-deaconu:
This code causes the 'Play received multiple times'
catcher to activate

Otherwise, Presentations continue playing past their Timeline expiry

To be more specific:
Presentation Play is called (isPlaying = true) > Presentation Done is called (isPlaying = true) > Presentation Play is called but PH play is not because isPlaying = true

Fix: 
Presentation Play is called > Presentation Done is called > Presentation Play is called > Presentation Done is called > 'Play received multiple times' catcher > Presentation Stop is called.


fjvallarino:
:+1:

PR #215, opened 2017-01-23
If the current item is PUD, interrupt any previous timers

alex-deaconu:
Verified the following setup:
Schedule Item 1 - 10 seconds duration, timeline within the Presentation set to not Play
Schedule Item 2 - Play Until Done, has duration longer than 10 seconds
Schedule Item 3 - anything

Noticed that the second item would only play for 10 seconds. The timer set for Item 1 would still be running (as it is not interrupted) while the second item (which is PUD) was playing. Once that timer expired, the PUD item would be switched and Item 3 would appear.

Fix resets timer for PUD items; in most circumstances this is not necessary but for this use case it seems to fix the issue.


fjvallarino:
@alex-deaconu looks safe. I'm wondering if we shouldn't be resetting the timers when running a PUD earlier in the code (just in case we're exiting on a previous condition).

Beyond that, :+1:

alex-deaconu:
@fjvallarino this PR fixes the specific issue with the timers as described, but there may be other underlying issues. I was actually thinking the same thing as you're suggesting. Will investigate further and see if it makes sense to interrupt the timer in a different spot. Thanks.

alex-deaconu:
@fjvallarino updated. Change is simple enough, but I had to make sure it doesn't break stuff elsewhere. Please review. Thanks!

fjvallarino:
:+1:

PR #217, opened 2017-01-25
If Presentation overwrites jquery version, jquery timers is missing

alex-deaconu:
Causes errors when rendering Widgets and receiving Ready event


alex-deaconu:
Fixes issue identified here: https://community.risevision.com/rise_vision_inc/topics/widgets-do-not-work-with-merged-html

Thought this was a very interesting variation to using Presentations. Figured it was something we should enable the developer to do so I am submitting a fix. To see it in action, you can preview here:
http://jquery-dependency-dot-rvaviewer-test.appspot.com?type=presentation&id=26709602-60bc-466b-8ba7-5f7575e8accb

@ezequielc @fjvallarino please review. Thanks.

fjvallarino:
@alex-deaconu doesn't it affect Viewer if those jQuery timers are not found? With this change it will clearly not "crash", but will the rest of Viewer work correctly?

alex-deaconu:
@fjvallarino this change is isolated to the Presentation iframe itself, so it wouldn't affect the whole Viewer app. 

The timers are used exclusively for the retry mechanism. Yes, it will affect that behaviour (if a Widget/Gadget/Image fails to load within 1 minute it will not retry) but only in the case where the JQuery object is overwritten by another JQuery library loaded in the Presentation itself.

An alternative to this quick fix is to refactor the code to use `setTimeout` instead of the jquery timers, but that would require a bit more work.

Thanks.

fjvallarino:
@alex-deaconu sounds good. In case you think it's a better idea to replace the jQuery timers with setTimeout, I can do it.

Otherwise, :+1:

alex-deaconu:
@fjvallarino in my opinion, this whole file (presentationScripts.js) needs a refactor. Inclusion in src-js, adding to the build process, unit tests. This is just a quick workaround for the uncaught errors preventing the content from loading.

fjvallarino:
@alex-deaconu just in case, I approve this again :+1:

PR #219, opened 2017-01-28
Improve/add integration tests

fjvallarino:
@alex-deaconu this is not actually a PR, but I created it since it should help for easier discussion. 

I changed most of the code to (try to) simplify the logic and help for easier understanding of the test cases (I don't think trying to compare both version with git diff will really help). Considering these are Integration tests, I tried to reduce mocking as much as possible, since we want to validate all the pieces work fine together (or, at least, as many pieces as we can run outside an actual browser). 

The first test case works fine, since it's actually simple. It tests that when running a schedule with two presentations, only the first one starts playing. Since it has two placeholders without specific timeline settings, both of them should start playing their first element (both placeholders are Playlists).

The second test case brings issues. It's basically the same as the first one, but it checks what happens when time passes (the intention is checking Presentation 2 does not start playing, since only enough time for finishing Presentation 1 will have passed; checks for Presentation 2 are intended for the next, unimplemented, test). I added a really basic for loop to have TimerController generate events (it works fine, all the callbacks are being called), which prompts PlaceholderController's _playNextItem to be called (as expected). The problem is, every single time _playNextItem is called, currentItem and nextItem are reset to -1 (not sure by what), although in the previous _playNextItem call both of them were increased to 0 by the time the function call was complete. Because of this, no other playlist items other than the first two (one from each placeholder) is played, although after 45 seconds the 3 items of placeholder 1 should have played, as well as the 2 items of placeholder 2.

-----

Note: The changes in karma.config.js are temporary (I will restore them before merging) and are there for making it easier to see tests results when running "npm run test" on a console, since it does not print the whole coverage information and I can see easily which tests failed. The same happens with marking the "integration" suit as the only one being ran.
Regarding the for loop for advancing time, I will implement it in a cleaner way; this version is just for testing.

@ezequielc FYI


fjvallarino:
@alex-deaconu I missed the important question, of course: do you know what could be causing this behavior?

ezequielc:
@fjvallarino @alex-deaconu Any conclusions on this? Thanks

fjvallarino:
@ezequielc @alex-deaconu no, currently changing the tests to rely more on EntryPoint's internals instead of handling the ScheduleController directly

fjvallarino:
@Rise-Vision/apps updated. In the end I only mocked ViewerDataProvider.retrieveData method (which I was not using, but was being ran by an internal timer). I left ViewerDataController unmocked, since I need some of its event handling in place (in particular. onDataReady).

alex-deaconu:
@fjvallarino looks good 👍 . Actually, this is a great improvement, the file is much more readable than before. Thanks!

fjvallarino:
@Rise-Vision/apps I imported the other two existing tests from Content.integration. I attempted to add some extra Video test cases, but was not able to make them work in a reasonable time.

I have not imported the ignored test, since I don't really know what the "known bug" mentioned there is. In case somebody knows what the issue is, I can try to import it.

Excluding that, I'm currently done with this set.

Please review!

src-js/Viewer/Content/Content.integration.tests.js
```diff
@@ -143,6 +143,43 @@ describe("integration:", () => {
...
+              }]
+            }
+          }
+        });
+
+        reportPlaceholdersReady(presentations);
+        expect(Utils.playCommand).not.to.have.been.called;
+
+        wait(100);
+        expect(Utils.playCommand).not.to.have.been.called;
```
alex-deaconu: @fjvallarino wondering if you can add a check that the content is actually showing? ie. check `window.setVisibility`. Thanks.

alex-deaconu:
@fjvallarino looks good. Just one suggestion re: emtpy Placeholder test. Thanks.

src-js/Viewer/Content/Content.integration.tests.js
fjvallarino: @alex-deaconu just added a validation. Let me know if that's the test you wanted to add

alex-deaconu:
@fjvallarino yup, looks good 👍 . Thanks!

PR #223, opened 2017-02-02
Beautify results

alex-deaconu:


alex-deaconu:
@fjvallarino please review. Thanks!

PR #211, opened 2017-01-18
Viewer Play Until Done mechanism fixes

alex-deaconu:
@ezequielc @fjvallarino the new branch is ready. As discussed, I've re-worked the commits in an order that makes more sense.

Thanks!

PR #224, opened 2017-02-03
Set up debug logging to clean up logs

alex-deaconu:
My suggestion is we merge this to #211 once testing is complete. 

To note, I considered making the `debug` variable part of the config, but since we also deploy staging with the `prod` environment selected, it would always be false. Just leaving it up to the developer to set it to `true` if we need to debug.


fjvallarino:
:+1:

PR #225, opened 2017-02-06
Fix issue with CacheController not returning result

alex-deaconu:
When in Preview, the callback is not called so the Video doesn't load


alex-deaconu:
Video Items do not play in Preview because of this. Found this issue while testing the current PR. It's unrelated, but occurs in Production. Figured it was a good idea to fix.

fjvallarino:
:+1:

PR #227, opened 2017-02-08
Only provide Presentation Done if item is PUD

alex-deaconu:
Don't provide PH done if Presentation done is not present

Some Unit test fixes


alex-deaconu:
The implementation is quite straight forward. Don't send the "Done" command if the Item is not Play Until Done.

fjvallarino:
:+1:

PR #228, opened 2017-02-09
Fix issue with tables created in the future

alex-deaconu:
Integrate extLogger into src-js
Fix Unit test conflicts


alex-deaconu:
The single difference between the 2 inserts is this: https://github.com/Rise-Vision/viewer/blob/master/src/main/webapp/scripts/ext-logger.js#L46

I would assume something in that code is causing the table name not to be generated correctly. So, removed that functionality and am setting the table name directly: https://github.com/Rise-Vision/viewer/pull/228/files#diff-39dc59a8db67ac190e491c9207d1d566R123

Also, lots of refactoring as it is now included in src-js. 

TODO: Remove Logger.logExternalMessage and Logger.logAppsEvent and update all the references in the code to use ExtLogger. Unfortunately that is a lot of work so I didn't do it as part of this card. Let me know if you think I should address.

fjvallarino:
:+1:

PR #229, opened 2017-02-14
Initialize EntryPoint module in Viewer.html

fjvallarino:


src-js/Viewer/EntryPoint.js
```diff
@@ -267,4 +263,6 @@ service.embedStop = function () {
   service.embedPause();
 };
 
+window.EntryPointInit = service.init;
```
alex-deaconu: @fjvallarino I don't think this is necessary. EntryPoint is available at `window.RiseVision.Viewer.EntryPoint`. Thanks.

src-js/Viewer/Logger.js
```diff
@@ -12,7 +12,10 @@ Logger.logDebug = (msg) => {
 };
 
 Logger.log = (msg, e) => {
-  window.writeToLog(msg);
+  if (window.writeToLog) {
+    window.writeToLog(msg);
+  }
```
alex-deaconu: @fjvallarino should perform a similar check in `logException` below. Thanks.

src/main/webapp/Viewer.html
```diff
@@ -369,6 +369,12 @@
 	</script>
 	<![endif]-->
 
+  <script type="text/javascript">
+    $(document).ready(function() {
+      window.EntryPointInit();
+    });
```
alex-deaconu: @fjvallarino curious, why `$(document.ready..)` and not `document.addEventListener("DOMContentLoaded"..)` as we originally had? Thanks.

fjvallarino: @alex-deaconu because the e2e tests (screenshot) were failing on Circle. It's probably caused by ES6 function syntax (()=>{}). I used the jQuery version because of compatibility, but DOMContentLoaded supposedly exists since Chrome v1...

alex-deaconu: @fjvallarino not sure I follow. Does the event work or does it fail? Obviously not using the es6 notation and just using `function() {}`. Thanks.

fjvallarino:
@Rise-Vision/apps updated

alex-deaconu:
@fjvallarino looks good. Thanks for the fixes. However, we cannot merge this until we release the current master version to 100% (eta Thursday). Thanks.

PR #233, opened 2017-02-24
Made changes to integrate viewer with player

rodrigopavezi:
Changed gitignore to keep viewer dist folder on github
Changed Viewer.html resource paths

@tejohnso @alex-deaconu please review. cheers


alex-deaconu:
@rodrigopavezi I can't say I'm a fan of the fact that the `dist` folder is now in the repo, but I would assume there are no alternatives we could consider to prevent that.

PR looks good 👍 . I would suggest coordinating with Support to perform a smoke test after the deployment; and I would advise against deploying this today, I would wait till next week.

Thanks!

rodrigopavezi:
@alex-deaconu we may not need to include the dist folder to git if we implement the GCS approach where viewer would be distributed through GCS and player would download it from there. @tejohnso  FYI. cheers

alex-deaconu:
@rodrigopavezi interesting. Curious, why not download the files from rvashow2.appspot.com/dist/*?

tejohnso:
@alex-deaconu What method would be used to get all the files from rvashow2.appspot.com/dist?  For GCS I think it's a simple `gsutil cp -r`

alex-deaconu:
@tejohnso not sure if there's a way to retrieve all the folder contents. Thanks.

PR #235, opened 2017-03-01
Include gsutil creds

tejohnso:
@fjvallarino please review


circle.yml
```diff
@@ -37,8 +37,11 @@ deployment:
...
     commands:
+      - git clone git@github.com:Rise-Vision/private-keys.git
+      - mv private-keys ..
+      - gcloud auth activate-service-account 452091732215@developer.gserviceaccount.com --key-file ../private-keys/storage-server/rva-media-library-ce0d2bd78b54.json
       - ./node_modules/.bin/cross-env NODE_ENV=prod npm run build
       - mvn appengine:update -Dappengine.version=r$(echo -n $((CIRCLE_BUILD_NUM%15))) -Dappengine.appId=rvashow2
       - mvn appengine:set_default_version -Dappengine.version=r$(echo -n $((CIRCLE_BUILD_NUM%15))) -Dappengine.appId=rvashow2
-      - create-local-viewer
-      - upload-local-viewer
+      - ./create-local-viewer
```
fjvallarino: @tejohnso should we use "sh ./create-local-viewer.sh" instead, since the file is probably not marked as executable?

tejohnso: @fjvallarino It's marked executable in the repo so I think it should work https://github.com/Rise-Vision/viewer/blob/master/create-local-viewer.sh

fjvallarino: @tejohnso what about the missing extension? I just tried on sh, bash and zsh and running the script without ".sh" failed for me.

tejohnso: @fjvallarino updated with extension

fjvallarino:
:+1:

PR #237, opened 2017-03-02
Use proxy-server to proxy makeRequest call in Offline Player

stulees:
- When document host is not *rvashow.appspot.com* , construct url to use our proxy server to call makeRequest with hardcoded url "http://rvashow.appspot.com/gagdets/makeRequest" . This is in order to get around cross domain request when Viewer is served locally from Offline Player


stulees:
@ezequielc @tejohnso @rodrigopavezi @fjvallarino  Please review, thanks

PR #238, opened 2017-03-03
Duplicate dependency dirs in main

tejohnso:
Viewer.html expects dependencies in the parent directory (eg:
../styles/viewer.css) while the presentation iframes expect the
dependencies in the same directory as Viewer.html.


stulees:
@tejohnso 👍

PR #239, opened 2017-03-10
Hardcode make request url for Offline Player with exception of localhost

stulees:


stulees:
@tejohnso Please review, thanks. @ezequielc @alex-deaconu FYI

tejohnso:
@stulees why would "localhost" be in the URL?

stulees:
@tejohnso The check for "localhost" is solely for enabling Viewer to work as normal when testing Viewer by running `npm run dev-server` because if we apply the hardcoded url in this situation then the requests are cross domain and fail.  I have tested this when running dev-server and it works.

tejohnso:
@stulees there's nothing there indicating that it's only a change that affects *npm run dev-server* and that's not an obvious scenario to consider in production code.  I think something like the following would help:

```
function isLocalDevServer() {
 return document.location.host.indexOf( "localhost" ) > -1;
}

var proxyHost = isLocalDevServer()  ? "//" + document.location.host : "http://rvashow.appspot.com";

var proxyUrl = proxyHost + "/gadgets/makeRequest";
```

stulees:
@tejohnso Good point, I've revised for clarity. Please review last commit, thanks

tejohnso:
:+1:

PR #241, opened 2017-03-17
Handle failure when parsing external data

tejohnso:
When offline, or on network failure, the parse will throw an exception
without this.

@fjvallarino @stulees please review


stulees:
@tejohnso 👍

PR #242, opened 2017-03-17
Notify player when ready for data

tejohnso:
@stulees @fjvallarino please review.  This resolves a sync issue where player sometimes sends the presentation data before viewer has registered to receive it.


fjvallarino:
:+1:

PR #244, opened 2017-04-28
Check subscription status

fjvallarino:
@Rise-Vision/apps not actually ready for review since build is failing and I need to add some unit testing, but meanwhile I'd like to get some feedback/suggestions about the implementation itself.


src-js/Viewer/Data/SubscriptionStatusProvider.js
```diff
@@ -0,0 +1,87 @@
...
+    cb(subscribed);
+  })
+  .fail((jqXHR, textStatus, error) => {
+    Logger.logException("SubscriptionStatusProvider._getAuthorizedStatus.fail", error, url, textStatus);
+    cb(false);
+  });
+};
+
+service.loadStatus = (companyId, productCode, cb) => {
+  _getSubscriptionStatus(companyId, productCode, (status, subscribed) => {
```
ezequielc: @fjvallarino I think you can skip the subscription status call and request authorized status only. It should cover what we need. Thanks

fjvallarino:
@Rise-Vision/apps updated

src-js/Viewer/Config/dev.js
```diff
@@ -3,4 +3,5 @@ module.exports = {
   APPS_URL: "https://apps-stage-0.risevision.com",
   PREVIEW_TRACKER_ID: "UA-82239-27",
   DISPLAY_TRACKER_ID: "UA-82239-31",
+  STORE_SERVER_URL: "https://store-dot-rvaserver2.appspot.com/"
```
alex-deaconu: @fjvallarino I think we should be using `store.rvacore-test.appspot.com` for dev & test. Thanks.

src-js/Viewer/Content/PlaylistItemController.js
```diff
@@ -296,6 +300,15 @@ const PlaylistItemController = function (placeholderInfo, playlistItem, phName,
     PresentationParser.addScriptTag(presentation, scriptTag);
   };
 
+  let _checkSubscriptionStatus = () => {
+    // Check subscription authorized status for embedded presentations
+    const companyId = window.RiseVision.Viewer.Data.ViewerDataInfo.getItemCompanyId();
+
+    SubscriptionStatusProvider.checkAuthorizationStatus(companyId, Global.EMBEDDED_PRESENTATIONS_PC, (authorized) => {
+      Logger.log("Subscription authorized status", authorized);
```
alex-deaconu: @fjvallarino I think in order to log to BQ you need to use the `logExternal` function. Also, I would suggest including the word Embedded Presentation in the log entry. Thanks.

src-js/Viewer/Data/SubscriptionStatusProvider.js
```diff
@@ -0,0 +1,114 @@
...
+  const cacheKey = companyId + productCode;
+
+  if (typeof cachedAuth[cacheKey] !== "undefined") {
+    Logger.log("Cached authorization - " + cachedAuth[cacheKey]);
+    cb(cachedAuth[cacheKey]);
+  }
+  else {
+    _getAuthorizedStatus(companyId, productCode, (authorized) => {
+      cachedAuth[cacheKey] = authorized;
+      cb(authorized);
```
alex-deaconu: @fjvallarino to confirm, from what I can tell this code will still call the API for every instance of an Embedded Presentation item, until a response is received and cached. My suggestion is you include a cached promise (I think it should be available via the es6 converter) or something of the sort to only call the API once, but respond to all callbacks. Thanks.

alex-deaconu:
@fjvallarino reviewed. Couple of comments. Thanks.

fjvallarino:
@alex-deaconu updated to address the comments

src-js/Viewer/Content/PlaylistItemController.js
```diff
@@ -296,6 +300,16 @@ const PlaylistItemController = function (placeholderInfo, playlistItem, phName,
     PresentationParser.addScriptTag(presentation, scriptTag);
   };
 
+  let _checkSubscriptionStatus = () => {
+    // Check subscription authorized status for embedded presentations
+    const companyId = window.RiseVision.Viewer.Data.ViewerDataInfo.getItemCompanyId();
```
alex-deaconu: @fjvallarino after some testing, it appears that the companyId is blank in the case of a Presentation Preview. It appears this line is incorrect: https://github.com/Rise-Vision/viewer/blob/master/src-js/Viewer/Data/ViewerDataInfo.js#L156 . `item.companyId` is always null.

To fix, I would suggest changing this to:
`itemCompanyId = content.presentations[index].companyId;`

Thanks.

src-js/Viewer/Content/PlaylistItemController.js
```diff
@@ -296,6 +300,16 @@ const PlaylistItemController = function (placeholderInfo, playlistItem, phName,
...
+
+    SubscriptionStatusProvider.checkAuthorizationStatus(companyId, Global.EMBEDDED_PRESENTATIONS_PC, (authorized) => {
+      Logger.log("Subscription authorized status for Embedded Presentation", authorized);
+      Logger.logExternalMessage("Subscription authorized status for Embedded Presentation", authorized);
```
alex-deaconu: @fjvallarino when `authorized` is false, the value logged is blank:
screen shot 2017-05-05 at 11 21 00 am . Maybe try logging `authorized.toString()` since the value is boolean. Thanks.

alex-deaconu:
@fjvallarino code looks complicated, but it works. However, tested the functionality and found a couple of potential issues. Thanks.

src-js/Viewer/Content/PlaylistItemController.js
ezequielc: @alex-deaconu @fjvallarino As far as I remember, we decided to not check authorization in Preview. I don't think we need to add logging in that case also. Thanks

alex-deaconu: @ezequielc I don't recall that discussion. Wouldn't that mean that users can run Previews on their Displays to get Embedded Presentations for free? Thanks.

alex-deaconu: @ezequielc @fjvallarino either way this is a bug; Presentation Previews would be getting a null Company Id in the case Widgets request it. Thanks.

ezequielc: @alex-deaconu Yes, it would, but I believe this is how Widgets work. 
Because once you make a call to Authorization API, the trial starts automatically (as mentioned here: https://help.risevision.com/developer/store-api/store-auth/store-start-trial-api). I don't think it should start a trial only by previewing. I will confirm Brian. Thanks

fjvallarino:
@alex-deaconu updated with two new commits

Agreed on the code complexity. Handling the single server call with callbacks is cumbersome.

alex-deaconu:
@fjvallarino looks good, thanks for fixing. I think the last thing we need to do is determine whether or not we need to call the API in Preview. @ezequielc any updates on that? Thanks.

ezequielc:
@alex-deaconu @fjvallarino I didn't get a conclusive answer on it, but Storage component checks for subscription in Preview, so I am ok with checking it too. We can review that in future if needed. Thanks!

alex-deaconu:
@fjvallarino PR looks good 👍 . Thanks!

PR #245, opened 2017-05-01
fix rewire error

settinghead:


settinghead:
@tejohnso Looks like the rewire issue can be resolved by upgrading the version of a plugin. Let me know if you still have any questions. Thanks.

settinghead:
@tejohnso I agree, it looks highly unstable. From my experience, `^` upgrades are not supposed to break things that are already running, but in reality not many module maintainers follow this strictly and I've seen things break randomly. I've had a few times where I was fed up and just put things into fixed versions.

PR #247, opened 2017-05-15
Limit rise cache ping attempts

tejohnso:
We have machines that aren't respecting the timer delay so I'm doubtful changing that is going to work.  Instead I added a sanity check on the number of attempts made.


src-js/Viewer/Cache/RiseCacheController.js
```diff
@@ -75,6 +76,13 @@ service.getCacheVideoUrl = function (url, extension, callback) {
 };
 
 service.pingCache = function () {
+  if (pingAttempt >= MAX_PING_ATTEMPTS) {
+    isActive = false;
+    _clearCallbacks();
```
stulees: @tejohnso I think it would be good to also call `clearTimeout(pingInterval);` to ensure no additional timer executes, which is consistent with code in `_pingResponseStatic` function.

tejohnso:
@stulees updated

PR #248, opened 2017-05-15
Adding messaging service module

stulees:


src-js/Viewer/Messaging.js
```diff
@@ -0,0 +1,73 @@
+const Logger = require("./Logger");
+
+const service = {};
+const handlers = [];
+
+let connection;
+
+service.init = (displayId) => {
+  const serverUrl = `https://display-messaging.risevision.com${(displayId) ? `?displayId=${displayId}` : ""}`;
```
tejohnso: If there's no displayid we shouldn't connect at all.  This should be handled with log and return.  We should probably external log since this would be a pretty strange scenario.

stulees:
@tejohnso Changes made and tests added, please review again, thanks

src-js/Viewer/Messaging.js
```diff
@@ -0,0 +1,80 @@
...
+
+  if (!window.Primus) {
+    Logger.log("primus.js was not loaded");
+    return;
+  }
+
+  if (!displayId) {
```
tejohnso: I think the ternary in line 9 can be removed now.

stulees:
@tejohnso Change made

PR #251, opened 2017-05-17
Enable all unit tests to run

alex-deaconu:
@tejohnso @stulees while trying to make a small modification to the Viewer, I've noticed that all unit tests were disabled. Upon enabling tests, the build now fails. 

Is this something you guys can address?

Thanks.


tejohnso:
@alex-deaconu see https://github.com/Rise-Vision/viewer/pull/252

PR #252, opened 2017-05-17
Re-enable all tests

tejohnso:
@alex-deaconu sorry about that, forgot a `.only` during yesterday's changes.


alex-deaconu:
@tejohnso looks good, thanks for fixing! 👍

PR #256, opened 2017-05-18
Log errors regardless of existence of error object

alex-deaconu:
To confirm, it seems that when an external `jQuery.getJSON` call fails (due to proxy, or invalid URL, etc), the `.fail(function(jqXHR, textStatus, error))` callback is called with an undefined `error` object. When we call `logException` with that undefined object the function quits and doesn't log anything.

Fixing this, however to note Viewer error count will go up in cases where these calls fail.

To confirm, this is just needed for logging Authorization API failures for Embedded Presentations subscription status checks. I could implement a local fix for this if we don't want other API calls to log.

@ezequielc @tejohnso @stulees please review, thoughts? Thanks!


alex-deaconu:
Makes sense. Thanks for the review!

PR #264, opened 2017-06-08
Add retry mechanism for authorization api call

alex-deaconu:
Store response in localStorage
If API call fails 3 times, attempt to use local storage value


alex-deaconu:

fjvallarino:
@alex-deaconu code looks good, although I'm not sure how to properly test it. Disabling the network in Chrome is tricky, since I either do it too early (Viewer did not load) or too late (the auth part has already taken place).

alex-deaconu:
@fjvallarino nice, thanks for the review & testing!

PR #265, opened 2017-06-15
Respect dupe id from msg svc

tejohnso:
Dupe id messages are ignored if connection is from the same machine.


src-js/Viewer/Machine-id.js
```diff
@@ -0,0 +1,27 @@
+let id = setMachineId();
+
+function setMachineId() {
+  if (id) { return id; }
+
+  try {
```
stulees: @tejohnso I think this try/catch can be omitted for `id = localStorage.getItem("machineId");`

tejohnso: If there's no localStorage we'll throw ReferenceError.  This just squelches that since there's nothing we would do with an error message.

stulees: @tejohnso In this initial try/catch, aren't you trying to retrieve the value of machine id if its been previously saved already? Using getItem("machineId")  is to retrieve it, I don't think localStorage.machineId will ever return a valid value.

tejohnso: @stulees yes, that's all I'm doing and avoiding errors if localStorage doesn't exist at all.  This way works as well as getItem and `localStorage["machineId"]` would also work.  Tested in viewer.

stulees: @tejohnso Okay thanks

PR #268, opened 2017-06-29
Show error message if Embedded Presentations not Subscribed

alex-deaconu:
- Block item from being shown until Subscription status is validated
When isError is set, isReady() returns false

When Subscription status is received, if true, isReady() will return true

If false, destroy element and render a text element with error message

Behaviour should be consistent if Item is ready or Subscription response is received first

- Fix issue where polling is done even after status is received
- Fix issue where Embedded Presentations containing other Embedded Presentations (3rd layer, or more) no longer check Subscription status (and automatically assume Parent Presentation performed the check)

- Some refactoring


alex-deaconu:
Testing URLs:

Subscribed: http://unsubscribed-dot-rvaviewer-test.appspot.com/?type=presentation&id=1787bbfc-cec6-4ba9-ab90-28cfbe58a85f

Bunch of Embedded Presentations in an expired Company: http://unsubscribed-dot-rvaviewer-test.appspot.com/?type=presentation&id=342a699f-0ce2-406b-9cfe-0f8db649dab9

Expired subscription with black background (middle right placeholder with the map as the first item): http://unsubscribed-dot-rvaviewer-test.appspot.com/?type=presentation&id=0a07f517-8506-4451-9998-b26f4fb26e25

As per the last link I should probably make the background of the "error" text white. Just wanted to confirm with you guys first however.

ezequielc:
@alex-deaconu I think you can use `text-shadow: 1px 1px #999;`. This is how it will look like:
screen shot 2017-06-29 at 6 41 27 pm
What do you think? THanks

alex-deaconu:
@ezequielc interesting. Added as per your suggestion, and it works as intended. However, I'm wondering if we should be concerned re: UX standards? @RV-Adrian any opinion on this? Thanks.

RV-Adrian:
@alex-deaconu @ezequielc @fjvallarino 

Played around with it a bit, ended up here:

image
image

Thoughts?

Inline style would be:
`  text-shadow: 1px 1px 1px rgba(0, 0, 0, .3);
    position: relative;
    padding: 14px;
    background-color: rgba(80, 80, 80, 0.8);
    top: 50%;
    transform: translateY(-50%);
    text-align: center;
    margin-right: 15px;
    color: white;
    font-family: sans-serif;
`

alex-deaconu:
Thanks @RV-Adrian . Updated as per your suggestion.

PR #276, opened 2017-08-08
Changed subscription status check to cache status on RC

rodrigopavezi:
@alex-deaconu @fjvallarino please review.

@tejohnso FYI.


src-js/Viewer/Data/SubscriptionStatusProvider.js
```diff
@@ -151,9 +151,12 @@ const _getAuthorizedStatusAttempt = (url, pingAttempt, cb) => {
...
 
   jQuery.getJSON(url, (response) => {
-    clearTimeout(pingInterval);
-
-    cb(response);
+    if (!(response && response.status === 202)) {
+      clearTimeout(pingInterval);
+      cb(response);
+    } else {
+      _getAuthorizedStatusAttempt(url, pingAttempt + 1, cb);
```
alex-deaconu: @rodrigopavezi wondering if we need to increment `pingAttempt` here. If we get 3 `202` statuses in a row (ie. API response is slow) we may surpass the `MAX_PING_ATTEMPTS` limit. Thoughts? Thanks.

src-js/Viewer/Data/SubscriptionStatusProvider.js
```diff
@@ -151,9 +151,12 @@ const _getAuthorizedStatusAttempt = (url, pingAttempt, cb) => {
   }, PING_TIME);
```
tejohnso: I think this timeout can be queued multiple times from line 150.  Is there a clearTimeout missing before line 150?

src-js/Viewer/Data/SubscriptionStatusProvider.js
rodrigopavezi: @alex-deaconu It can be because I think RC won't get stuck on 202 anyway. I will remove it.

src-js/Viewer/Data/SubscriptionStatusProvider.js
rodrigopavezi: @tejohnso makes sense. cheers

rodrigopavezi:
@alex-deaconu @tejohnso please review latest commit. cheers

tejohnso:
@rodrigopavezi How about an integration test that loads RC, then starts Viewer, then confirms they're working together as expected.

rodrigopavezi:
@tejohnso good point, I think we can do that on player e2e as RC is already started. Thoughts? cheers

tejohnso:
@rodrigopavezi sounds good.

PR #279, opened 2017-09-13
Fixed embedded presentation not showing in preview and on player

rodrigopavezi:
@alex-deaconu @tejohnso I had to change a bit my previous implementation of caching store embedded subscription into RC. I implemented the getCacheUrl using this async method _getIsActiveAsync which would hang for almost ever when RC was not running. So I changed to call
RiseVision.Viewer.Cache.RiseCacheController.getCacheUrl only if the presentation is playing on a display as I added a check for RiseVision.Viewer.EntryPoint.isDisplay().

I have also removed the encoding of the url parameter when we have RC v2 running as we have removed all decoding from newest RC.

@stulees FYI.
Please review. cheers


stulees:
@rodrigopavezi 👍 for the cache controller changes. I will defer to @alex-deaconu and @tejohnso for the subscription status provider changes,  cheers.

ezequielc:
@rodrigopavezi I am wondering if it would be worth to add a basic embedded presentation e2e test, to prevent similar issues in the future. Would that be easy to do? Thanks

rodrigopavezi:
@ezequielc yeap, agreed. I am doing it right now. cheers

ezequielc:
@rodrigopavezi Awesome! Thanks

tejohnso:
@rodrigopavezi looks good, assuming video and embedded tests are both using the same screenshot deliberately.

PR #283, opened 2017-11-10
Fix issue where Schedule URL is not showing

alex-deaconu:
Check for null presentation object
Additional checks

Updated the e2e schedule to include an additional item (a URL). This is the 4th item. The 3rd item (viewer-e2e-multi-right) is disabled from showing via timeline.

@ezequielc @fjvallarino @tejohnso please review. Thanks!


src-js/Viewer/Renderer/PresentationRenderer.tests.js
```diff
@@ -0,0 +1,117 @@
...
+      });
+
+      it("should show pointer", () => {
+        isDisplay = false;
+        PresentationRenderer.render({ type: "presentation" }, presentation, "htmlName", "containerName");
+
+        expect(window.createPresentation.args[0][9]).to.be.false;
+      });
+
+      it("should show pointer", () => {
```
fjvallarino: @alex-deaconu should this one be named differently, in order to avoid clashing with the previous test?

alex-deaconu:
@fjvallarino good point, updated. Thanks!

src-js/Viewer/Renderer/PresentationRenderer.tests.js
```diff
@@ -0,0 +1,117 @@
...
+        PresentationRenderer.render({ type: "presentation" }, presentation, "htmlName", "containerName");
+
+        expect(window.createPresentation.args[0][9]).to.be.false;
+      });
+
+      it("should not show pointer on a display", () => {
+        presentation.hidePointer = false;
+        PresentationRenderer.render({ type: "presentation" }, presentation, "htmlName", "containerName");
+
+        expect(window.createPresentation.args[0][9]).to.be.false;
```
fjvallarino: @alex-deaconu should this assertion test for true in this case? Argument index 9 on createPresentation is called hidePointer (https://github.com/Rise-Vision/viewer/blob/fbcec548f839cbd3aeedbe2bb6f2abdc160d5876/src/main/webapp/scripts/viewerScripts.js#L169)

alex-deaconu:
@fjvallarino you're right; was confused regarding the variable. Should be more clear now. Thanks.

PR #284, opened 2017-12-12
Don't log to player config if no display id

stulees:
Details [here](https://trello.com/c/U99VIPHa/3620-2-viewer-doesnt-log-to-player-config-if-no-display-id)


alex-deaconu:
@stulees I would check with `EntryPoint.isDisplay()` rather than the existence of the Display Id. Embedded Presentations will return a Display Id, but isDisplay would be false. We most likely want to ignore logging for Embedded Presentations as well. Thanks.

tejohnso:
We definitely need a displayId in the player config table so I think we should keep the abandon if no displayId check.  If we can also add the isDisplay() check and remove those that's good too. At least for player config.

But I wonder if we should add this for the viewer events and apps events functions as well.  Viewer events has over 600,000 records with no display id for yesterday alone.

@alex-deaconu do you guys have use for log entries with no display id in the viewer or apps events tables?

alex-deaconu:
@tejohnso the amount of entries is related to this issue (https://trello.com/c/VFEnAmnC/2852-viewer-investigate-cost-increase-for-oct-nov-the-app-engine-project-1), which we're moving forward to resolve with the user. They generate ~200k viewer requests per day.

So I'm assuming each Viewer Preview logs ~3 events. Once these requests stop this number should be manageable, don't believe it's something we should concern ourselves with at this point. 

Thanks.

stulees:
@alex-deaconu @tejohnso Added the isDisplay() check along with valid displayId check, please review, cheers.

PR #285, opened 2018-01-17
Open connection to Local Messaging service and provide global API to …

stulees:
…write and receive messages


src-js/Viewer/LocalMessaging.js
```diff
@@ -0,0 +1,95 @@
...
+const INITIAL_CONNECTION_TIMEOUT = 20 * Global.ONE_SECOND_DELAY;
+const service = {};
+const messageHandlers = [];
+
+let connection;
+let connectivity = false;
+let initialConnectionTimer = null;
+
+service.init = (embeddedViewer, preview, cb) => {
+  if (window.disableViewerContentFetch) { return cb(); }
```
tejohnso: I think this should be !window.disableViewerContentFetch. 
If that field is on the window object it means rise-player-electron started the Viewer which is what we want. 

For the Messaging module, we exit if player started the Viewer, because Player handles the messaging in that case.

stulees: @tejohnso Thanks for catching. Change made, see latest commit

PR #292, opened 2018-03-15
Set domain of document origin to relax same-origin policy for widgets

stulees:
- The same will be set in Widgets
- This is to enable widgets to access `window.top` and workaround browser cross-origin restrictions


src/main/webapp/Viewer.html
```diff
@@ -400,7 +400,11 @@
 
   <script type="text/javascript">
    if (document.domain.indexOf("localhost") === -1) {
-     document.domain = "risevision.com";
+     try {
```
andrecardoso: Why wrapping it in try/catch? In which situation is going to throw?

stulees:
@tejohnso @andrecardoso FYI, I tested with this staged branch of Viewer in Electron Player and Viewer broke because it was not handling the DOM exception error from setting document.domain due to "risevision.com" not being an accepted suffix of current document domain (viewer-domain-dot-rvaviewer-test.appspot.com). This would happen also with production Viewer in Electron (rvashow2.appspot.com). 

I've now wrapped it in a try/catch to resolve it and I validated.

src/main/webapp/Viewer.html
stulees: @andrecardoso See my other comment, thanks

PR #323, opened 2019-04-24
Deprecate stop command; use pause instead

alex-deaconu:
Remove unused boolean variables for play/pause
Functionality no longer applies


alex-deaconu:
The following references in the code imply the `stop` command is not very consistently implemented, and `pause` is used as a fallback:
https://github.com/Rise-Vision/viewer/blob/master/src-js/Viewer/EntryPoint.js#L286
https://github.com/Rise-Vision/viewer/blob/master/src/main/webapp/scripts/videoScripts.js#L185
https://github.com/Rise-Vision/viewer/blob/master/src-js/Viewer/Content/ScheduleController.js#L285

Removing this command should not cause any issues but will reduce confusion in the logic.

Unlike the other PR which is for an issue fix, I think this one we can stagger release for good measure.

@ezequielc @Rise-Vision/delivery please review. Thanks!

tejohnso:
@alex-deaconu Can you clarify:

1. How you know that nothing is expecting 'stopCommand' to be triggered
2. Why this is being done now, and whether it can wait until after player upgrade

alex-deaconu:
@tejohnso 
1. `play` and `pause` are used arbitrarily throughout the Viewer code. If a widget/gadget expects one and not the other it would not function correctly in some circumstances. Most actually implement one and link the function to the other, I don't know of any instances where their implementation is different or missing.

2. While working on the other fix, I had some trouble identifying why one is being called vs the other. I looked into it and saw that it's arbitrary, and confusing. This is just a simplification. It can be released whenever, but it doesn't have to be released at all.

Thanks.

PR #325, opened 2019-06-24
Send current Schedule Item to player

alex-deaconu:
This is used for Uptime Logging. It basically allows the Player to determine which HTML Template is Playing. The `item` object is pre-populated with the information the Player needs as per Player changes in separate PRs.

To confirm, there are two general events that trigger this. A new item playing, indicated by
`presentations[currentItem].play();`

and a Schedule Timeline or Playlist Item Timeline setting "failure", where a Black screen is shown:
`EntryPoint.showBlackScreen(true);`

This PR is very similar to the Player changes that relate to this.

Finally, Local Messaging is used to relay this to the Player. Having said this, the change can go out regardless of the Player listening to the event or not, ie. we can deploy this without having Player support for it as the message will simply be ignored.

@ezequielc @tejohnso @andrecardoso please review. Thanks!


alex-deaconu:
@ezequielc good question. 

That use case is the Viewer sending a Restart command to an item that has sent Done. So no, I don't think it's necessary to send a notification in that case.

Thanks.

src-js/Viewer/Player/PlayerMessaging.js
```diff
@@ -0,0 +1,13 @@
+const LocalMessaging = require("../LocalMessaging");
+
+const service = {};
+
+service.sendCurrentScheduleItem = function (item) {
+  const msMessage = {
+    topic: "playing-item",
+    item
```
andrecardoso: I think it is safe to use this shorthand syntax, it has been available [since Chrome 47](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Browser_compatibility), but I just want to confirm if you think it is OK to ship it. Thanks

andrecardoso:
@olegrise, please take a look. Thanks!

src-js/Viewer/Player/PlayerMessaging.js
alex-deaconu: @andrecardoso this change was a result of `gulp beautify`, ie. autocorrected. Sorry for not being clear on that but the spacing changes are from that as well. Thanks.

andrecardoso: OK, so I assume it is good to go. Thanks

PR #326, opened 2019-06-28
Add PUD for HTML Templates

olegrise:
@andrecardoso @alex-deaconu please review. Also see the corresponding PR in Common Template https://github.com/Rise-Vision/common-template/pull/66.
I used postMessage instead of local messaging because I think it's a better option for this case - simpler and more reliable.


alex-deaconu:
@olegrise looks good 👍 . My only concern is lack of test coverage for this and the https://github.com/Rise-Vision/common-template changes. Thanks.

olegrise:
@alex-deaconu Thanks for the feedback. We are going to improve test coverage in `common-template` in the next couple of days. Thanks.

PR #327, opened 2019-07-05
Send rise-presentation-{play/stop} events

andrecardoso:


src-js/Viewer/Content/PresentationController.js
```diff
@@ -273,9 +273,18 @@ const PresentationController = function (item, htmlName, containerName,
...
+    const presentationIframe = document.getElementById("iFrame_" + htmlName);
+    if (presentationIframe) {
+      presentationIframe.contentWindow.postMessage({ topic }, "*");
+    }
+  };
+
   factory.play = () => {
     Logger.logDebug(`Presentation ${htmlName} ${EntryPoint.isEmbed() ? "(Embedded) " : ""}- Play!`);
 
+    _sendPresentationEvent("rise-presentation-start");
```
alex-deaconu: @andrecardoso curious re consistency of naming. Everywhere so far this event has been called `play` (viewer, widgets, etc), now we're renaming to `start`. Thoughts? Thanks.

src-js/Viewer/Content/PresentationController.js
```diff
@@ -273,9 +273,18 @@ const PresentationController = function (item, htmlName, containerName,
     }
   };
 
+  const _sendPresentationEvent = (topic) => {
+    const presentationIframe = document.getElementById("iFrame_" + htmlName);
+    if (presentationIframe) {
+      presentationIframe.contentWindow.postMessage({ topic }, "*");
```
alex-deaconu: @andrecardoso I would be inclined to create some sort of util class for Presentation Template communication, include this and the functionality above for PUD. The `Renderer/` folder would be a great spot for this. Some tests would be nice as well. Thanks!

alex-deaconu:
@andrecardoso looks good, some suggestions inline.

One thing I'd like to note is that the Viewer is blind as to whether it's dealing with a URL or a Template. This is not ideal, and probably not in the scope of this release, but maybe we can think of ways to be more transparent as to the type of content being rendered. 

Thanks.

src-js/Viewer/Content/PresentationController.js
andrecardoso: I'll change it to play

andrecardoso:
[(2) Finish rise-presentation-play/stop events implementation in viewer and common-template](https://trello.com/c/n2iNodyj/162-2-finish-rise-presentation-play-stop-events-implementation-in-viewer-and-common-template)

src-js/Viewer/Content/PresentationController.js
andrecardoso: @alex-deaconu I created the class as suggested, but I'm not so sure it is better now. Can you please re-review? Thanks.

alex-deaconu:
@andrecardoso looks good 👍 . Thanks!

PR #328, opened 2019-07-12
Fix HTML template done

andrecardoso:
- Only call presentation done if the template is playing


andrecardoso:
[(2) Viewer is transitioning items when a non-visible schedule template item sends the report-template-done event](https://trello.com/c/RR8oAGF3/171-2-viewer-is-transitioning-items-when-a-non-visible-schedule-template-item-sends-the-report-template-done-event)

src-js/Viewer/Content/PresentationController.js
```diff
@@ -51,7 +51,11 @@ const PresentationController = function (item, htmlName, containerName,
   let iframeName;
   let presentation;
   const placeholders = [];
-  const htmlTemplateMessaging = new HTMLTemplateMessaging(onPresentationDone, htmlName);
+  const htmlTemplateMessaging = new HTMLTemplateMessaging(() => {
+    if (isPlaying && typeof onPresentationDone === "function") {
+      onPresentationDone();
+    }
```
olegrise: @andrecardoso this is hard to read and probably not ideal for unit testing. What do you think about moving this block of code 
```
() => {
    if (isPlaying && typeof onPresentationDone === "function") {
      onPresentationDone();
    }
}
```
into a separate method and call it something like `_onPresentationDoneIfPlaying`?

alex-deaconu:
General question, shouldn't the content be `stopped` if the Template is not playing? ie. this would be an indication of something not working correctly with the stop event. Thanks.

src-js/Viewer/Content/PresentationController.js
alex-deaconu: Curious, why check for a `typeof function` rather than be consistent with exiting logic here https://github.com/Rise-Vision/viewer/blob/master/src-js/Viewer/Content/PresentationController.js#L257? Thanks.

andrecardoso: > General question, shouldn't the content be `stopped` if the Template is not playing? ie. this would be an indication of something not working correctly with the stop event. Thanks.

@alex-deaconu you are right. I caught this while I was testing custom PUD and there was no handling to `"rise-presentation-stop"`. That's why I created this PR and also this for [the template](https://github.com/Rise-Vision/html-template-library/pull/226) and this other for  [common-template](https://github.com/Rise-Vision/common-template/pull/74)

Thanks.

alex-deaconu: @andrecardoso sounds good 👍 . Thanks!

alex-deaconu:
@andrecardoso looks good to me, thanks for addressing!

PR #360, opened 2020-04-09
Don't initialize messaging or GCS for sharedschedule

alex-deaconu:
## Description
Don't initialize messaging or GCS for sharedschedule

Refactored so each service uses EntryPoint
values rather than function parameters

Calling dom functions rather than imports
to remove EntryPoint circular depenency

## Motivation and Context
We don't want messaging to initialize for the Shared Schedules

## How Has This Been Tested?
Updated unit tests.
Tested both Display, Preview and Shared Schedule scenarios to ensure the services are or are not initialized as expected.

## Release Plan:
- As the Submitter, upon requesting review of this pull request, I confirm that the [Release Checklist](https://help.risevision.com/hc/en-us/articles/360031119991) has been completed. 
- As the Reviewer, upon approving the changes in this PR, I confirm I have reviewed and I agree that the [Release Checklist](https://help.risevision.com/hc/en-us/articles/360031119991) has been completed

#### Release Checklist Items Skipped?
No


alex-deaconu:
To note, this PR is opened to the sharedschedules branch, as it is dependent on the creation of the new type. I will update to merge to master as I think both changes should be released separately.

@ezequielc please review. Thanks!

